PR #589, opened 2022-11-08
Remove ingress API from the repo

dati18:
Remove ingress API


rosalieper:
I dont think what we want is delete the wikibase-ingress chart. I think we probably just want to edit this line https://github.com/wmde/wbaas-deploy/blob/main/charts/wikibase-ingress/templates/ingress.yaml#L1 and use `networking.k8s.io/v1` instead of `extensions/v1beta1` like it is done here `https://github.com/wbstack/charts/pull/104/files#diff-c57f9efc1032efeb12558d54f1e0124d7c9a95e573e9eeb114d2a4e3a9b7069dL3`

PR #590, opened 2022-11-08
k8s: local/staging migrate to ingress-nginx

tarrow:


deer-wmde:
not entirely sure on what to look out for here but it seems to work on my local cluster :+1:

deer-wmde (approved):

deer-wmde:
To get the monitoring module code right I will deploy this to staging so wen can peek at the logs

PR #596, opened 2022-11-22
tf: upgrade error rate to new ingress name

tarrow:


deer-wmde (approved):
looks great - right now I'm working on a PR for getting the changes from https://github.com/wmde/wbaas-deploy/pull/590/ into the prod conf - I think getting that to prod and making sure it works would be the next step as this would "break" the prod error dashboard right now

update: here it is https://github.com/wmde/wbaas-deploy/pull/597

PR #625, opened 2022-12-16
staging & local: deploy api chart 0.20.1

deer-wmde:
Depends on https://github.com/wbstack/charts/pull/110


deer-wmde:
changed to draft because [the actual mail recipient address](https://github.com/wmde/wbaas-deploy/pull/625/files#diff-bb11be128453a5ee0b102c8feaa9890809f7f925ac2dee60b620b1c351fdeb65R77) still needs to be adjusted

PR #641, opened 2023-01-04
docs: Update key names for recaptcha environment variables

AndrewKostka:


tarrow (approved):
Looks good to me! Although we might rather at some point adjust the variable names anyway to remove the `dev`

m90:
Another helpful thing I've encountered before is having a `terraform.tfvars.example` as a template you can copy over and populate.

tarrow:
yeah, that sounds good! Feel free to go ahead and create one if you like :)

PR #644, opened 2023-01-04
docs: add document on accessing gke clusters from local machines

m90:
As discussed in Mattermost today, this documents collects the steps I needed to take when setting up GKE cluster access from my local machine today.


doc/gke-cluster-access.md
```diff
@@ -0,0 +1,74 @@
...
+gcloud container clusters get-credentials wbaas-3 --zone europe-west3-a
+```
+
+## Using the contexts
+
+To use these contexts, use `kubectl config set`.
+E.g. in order to point your `kubectl` at staging, run:
+
+```
+kubectl config set-context gke_wikibase-cloud_europe-west3-a_wbaas-2
```
tarrow, line 63:
```suggestion
kubectl config use-context gke_wikibase-cloud_europe-west3-a_wbaas-2
```

snipped from `kubectl config --help`:
```
get-contexts      Describe one or many contexts
  get-users         Display users defined in the kubeconfig
  rename-context    Rename a context from the kubeconfig file
  set               Set an individual value in a kubeconfig file
  set-cluster       Set a cluster entry in kubeconfig
  set-context       Set a context entry in kubeconfig
  set-credentials   Set a user entry in kubeconfig
  unset             Unset an individual value in a kubeconfig file
  use-context       Set the current-context in a kubeconfig file
```

set sounds like what you want at a glance but it it's seemingly more akin to "create context". `use` is the command for switching between contexts

tarrow (commented):
Maybe this should also remove `prod-env.md` since that appears similar but now wildly outdated. Fixing this isn't a blocker though; it can always be removed in a followup PR by you or someone else :)

m90:
@tarrow Updated this according to your feedback,

tarrow (approved):

PR #652, opened 2023-01-10
refactor(helm): allow for loading values from common directory for all environments

m90:
This PR is preparation for DRYing up our helmfile usage as described in: https://github.com/helmfile/helmfile/blob/main/docs/writing-helmfile.md#release-template--conventional-directory-structure

It does multiple things:
- Define a default `release` that all other `release` definitions inherit from. This default loads values from __both__ a `common` directory as well as the per-env directory. If a file is not found, a warning is printed, but value resolution will continue. The per-env values take precendence, objects are deep-merged, arrays are not merged, but replaced
- Rename values files so that they all match the `<release>.values.yaml.gotmpl` pattern and can be picked up automatically
- Remove the unused values for `local/nginx-ingress`. `ingress-nginx` is what is currently in use.

Currently, __there are no common values defined__ so this PR is a no-op when it comes to diffing and applying.

I tested diffing this against local and staging where no differences could be found as expected.


k8s/helmfile/helmfile.yaml
```diff
@@ -75,8 +82,7 @@ releases:
     installed: {{ ne .Environment.Name "local" | toYaml }}
     namespace: default
     chart: ./../../charts/certificates
-    values:
-      - "env/{{ .Environment.Name }}/certificates.values.yaml.gotmpl"
+    <<: *default_release
```
m90, line 85: The documentation says `inherit` should exist, but at least on my local setup, it does not work and I will see the following error:

```
in ./helmfile.yaml: failed to read helmfile.yaml: reading document at index 1: yaml: unmarshal errors:
  line 96: field inherit not found in type state.ReleaseSpec
```

deer-wmde: that's good to know - I think the YAML anchors are fine for now. If you want feel free to leave a comment in the template section for this so we have a hint to use inheritance in the future

deer-wmde (commented):
I think this is awesome, great job!

I'm still a little bit unsure with the `missingFileHandler` feature, because the behaviour without it brings a nice failsafe with it. But since we want to introduce these common files anyway real soon I don't think it's a 'real' issue. Also this way the configuration is already cleaner and backwards compatible so to speak, so that's also nice.

I will try it out more in-depth tomorrow on my machine but it looks really good

m90:
> I'm still a little bit unsure with the missingFileHandler feature, because the behaviour without it brings a nice failsafe with it. But since we want to introduce these common files anyway real soon I don't think it's a 'real' issue. 

In case it makes you feel safer about merging and releasing this, we could also use the branch of this PR as the base branch for all PRs that actually deduplicate values, then merge this PR to `main` once __all values__ have been migrated. This probably forces me to rebase this branch a couple of times, but I guess that's ok too.

I'd be fine with both approaches in any case.

m90:
> I'm still a little bit unsure with the missingFileHandler feature, because the behaviour without it brings a nice failsafe with it. But since we want to introduce these common files anyway real soon I don't think it's a 'real' issue.

Another thing we could do is set the `missingFileHandler` back to `Error` and create an empty file in common for each entry.

This would give us:
- pro: everything works the same way everywhere
- pro: deployments fail early in case you do something wrong
- con: lots of empty files for no obvious reason, which might be confusing in some situations

deer-wmde:
Thanks for exploring the options here, I think for now it's okay to proceed with this as it is and once we have migrated all common config we can figure out how we want this to behave

deer-wmde (approved):

PR #656, opened 2023-01-12
refactor(helm): elasticsearch values are repeated across envs

m90:

This PR removes the duplicated configuration of elasticsearch and uses the same common config for all environments.

This state diffs cleanly against local, staging and production.

### DIFF
--- a/k8s/helmfile/env/common/elasticsearch.values.yaml.gotmpl
+++ b/k8s/helmfile/env/common/elasticsearch.values.yaml.gotmpl
@@ -1,0 +1,23 @@
+image: "wikibase/elasticsearch"
+imageTag: "6.8.23-wmde.6"
+imagePullPolicy: Always
+
+replicas: 3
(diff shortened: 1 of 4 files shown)


k8s/helmfile/env/common/elasticsearch.values.yaml.gotmpl
```diff
@@ -0,0 +1,23 @@
+image: "wikibase/elasticsearch"
+imageTag: "6.8.23-wmde.6"
```
m90, line 2: Should we repeat the version number for each env so we can have an easier rolling upgrade?

deer-wmde: yes please :)

m90: Done so.

deer-wmde (approved):

PR #657, opened 2023-01-12
refactor(helm): redis values are repeated across envs

m90:

This PR removes the duplicated configuration of redis and uses the same common config for all environments.

This state diffs cleanly against local, staging and production.

### DIFF
--- a/k8s/helmfile/env/common/redis.values.yaml.gotmpl
+++ b/k8s/helmfile/env/common/redis.values.yaml.gotmpl
@@ -1,0 +1,27 @@
+redisPort: 6379
+architecture: "replication"
+sentinel:
+  enabled: false
+auth:
+  enabled: true
+  existingSecret: redis-password
+  existingSecretPasswordKey: password
+
+master:
+  persistence:
+    enabled: true
+    path: /data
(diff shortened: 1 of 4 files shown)


k8s/helmfile/env/common/redis.values.yaml.gotmpl
```diff
@@ -0,0 +1,27 @@
...
+  existingSecretPasswordKey: password
+
+master:
+  persistence:
+    enabled: true
+    path: /data
+    subPath: ""
+    accessModes:
+      - ReadWriteOnce
+    size: 1Gi
```
m90, line 17: I wonder if such very specific values that are also somehow tied to physical resources should __never__ go into `common`, even if all environments use the same value.

deer-wmde: good thought, I think a sensible fallback doesn't hurt though for now. If the values would vary wildly (which could be tied to major cost differences) I'd agree and propose to remove them from `common` as soon as they differ, but for now personally I'd say we're good

deer-wmde (approved):

PR #667, opened 2023-01-13
 refactor (helm): ui values are repeated across envs

deer-wmde:

Refactors the YAML value files for UI deployments and adds a `common` file with shared values.

diffs are clean for staging & prod.

### DIFF
--- a/k8s/helmfile/env/common/ui.values.yaml.gotmpl
+++ b/k8s/helmfile/env/common/ui.values.yaml.gotmpl
@@ -1,0 +1,24 @@
+resources:
+  requests:
+    cpu: 1m
+    memory: 6Mi
+  limits:
+    cpu: 10m
+    memory: 20Mi
+ingress:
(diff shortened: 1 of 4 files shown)


k8s/helmfile/env/common/ui.values.yaml.gotmpl
```diff
@@ -0,0 +1,24 @@
+resources:
```
m90, line 1: This is fine right now, but I think we might want to discuss as a team "what goes where" and whether we even want to specify resources in common.

m90 (approved):
:broom:

k8s/helmfile/env/common/ui.values.yaml.gotmpl
deer-wmde: I agree. We could aim to create a new [ADR](/doc/adr) after that discussion

PR #671, opened 2023-01-16
refactor(helm): sql values are repeated across envs

m90:

This PR removes the duplicated configuration of api and uses the same common config for all environments.

This state diffs cleanly against local and staging. In production, a missing command is added to the db's init script:

```diff
default, sql-mariadb-primary-init-scripts, ConfigMap (v1) has changed:                                                                                                                                     
...                                                                                                                                                                                                        
      helm.sh/chart: mariadb-10.5.0                                                                                                                                                                        
      app.kubernetes.io/instance: sql                                                                                                                                                                      
      app.kubernetes.io/managed-by: Helm                                                                                                                                                                   
      app.kubernetes.io/component: primary                                                                                                                                                                 
  data:                                                                                                                                                                                                    
-   init-primary-only.sh: |-                                                                                                                                                                               
+   init-primary-only.sh: |                                                                                                                                                                                
      #!/bin/sh                                                                                                                                                                                            
      if [[ $(hostname) == *primary* ]]; then                                                                                                                                                              
          echo "Primary node, executing"                                                                                                                                                                   
      else                                                                                                                                                                                                 
          echo "NOT Primary node, EXITING"
          exit
      fi
+   
+     # Backup manager / wbaas-backup
+     echo "CREATE USER IF NOT EXISTS 'backup-manager'@'%' IDENTIFIED BY '${SQL_INIT_PASSWORD_BACKUPS}';" >> /tmp/sql-init-cmds-001.txt
+     echo "ALTER USER 'backup-manager'@'%' IDENTIFIED BY '${SQL_INIT_PASSWORD_BACKUPS}';" >> /tmp/sql-init-cmds-001.txt
     
      # https://docs.pingcap.com/tidb/v4.0/mydumper-overview#required-privileges
      echo "GRANT SELECT ON *.* TO 'backup-manager'@'%';" >> /tmp/sql-init-cmds-001.txt
      echo "GRANT RELOAD ON *.* TO 'backup-manager'@'%';" >> /tmp/sql-init-cmds-001.txt
      echo "GRANT LOCK TABLES ON *.* TO 'backup-manager'@'%';" >> /tmp/sql-init-cmds-001.txt
...
```

Looking at the source code for the underlying image https://github.com/bitnami/containers/blob/main/bitnami/mariadb/10.9/debian-11/rootfs/opt/bitnami/scripts/libmariadb.sh#L342 this should never run though (the database is already initialized) so this is basically a no-op for bookkeeping purposes.

### DIFF
--- a/k8s/helmfile/env/common/sql.values.yaml.gotmpl
+++ b/k8s/helmfile/env/common/sql.values.yaml.gotmpl
@@ -1,0 +1,159 @@
+architecture: replication
+
+auth:
+  forcePassword: true
+  existingSecret: sql-secrets-passwords
+
+primary:
+  persistence:
+    enabled: true
+    size: {{ .Values.services.sql.storageSize | quote }}
+  extraEnvVarsSecret: sql-secrets-init-passwords
+  configuration: |-
+    [mysqld]
+    skip-name-resolve
+    explicit_defaults_for_timestamp
+    basedir=/opt/bitnami/mariadb
+    plugin_dir=/opt/bitnami/mariadb/plugin
+    port=3306
+    socket=/opt/bitnami/mariadb/tmp/mysql.sock
+    tmpdir=/opt/bitnami/mariadb/tmp
+    max_allowed_packet=16M
+    bind-address=0.0.0.0
+    pid-file=/opt/bitnami/mariadb/tmp/mysqld.pid
+    log-error=/opt/bitnami/mariadb/logs/mysqld.log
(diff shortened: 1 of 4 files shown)


deer-wmde:
FYI about the difference you catched here: the production DBs were already set up when we introduced this change I think, so we created that user for the backups manually there.

> so this is basically a no-op for bookkeeping purposes.

indeed!

deer-wmde (approved):

PR #673, opened 2023-01-17
DNM: Example of releases array being concatenated

tarrow:
Note that this *isn't* really useful because now helmfile will error with "two releases both named sql"


m90:
Closing this as stale.

PR #674, opened 2023-01-17
docs(adr): describe "production as base case" setup for helmfile values

m90:
Ticket [T326643](https://phabricator.wikimedia.org/T326643).

As discussed earlier, this adds an ADR about how to handle Helmfile values in the future.

I also added a script that creates a TOC for us automatically.

### DIFF
--- a/doc/adr/0006-helmfile-base-values.md
+++ b/doc/adr/0006-helmfile-base-values.md
@@ -39,11 +39,11 @@ Such a layout would look like this:
 .
 └── env
     ├── local
-    │    └── app.values.yml.gotmpl <-- local overrides only, plus image tag if applicable
+    │    └── app.values.yaml.gotmpl <-- local overrides only, plus image tag if applicable
     ├── production
-    │    └── app.values.yml.gotmpl <-- base case, containing a full configuration
+    │    └── app.values.yaml.gotmpl <-- base case, containing a full configuration
     └── staging
-        └── app.values.yml.gotmpl <-- staging overrides only, plus image tag if applicable
+        └── app.values.yaml.gotmpl <-- staging overrides only, plus image tag if applicable
 ```
 
 Example values for a hypothetical app could look like this:
--- a/doc/adr/0006-helmfile-base-values.md
+++ b/doc/adr/0006-helmfile-base-values.md
@@ -39,11 +39,11 @@ Such a layout would look like this:
 .
 └── env
     ├── local
-    │    └── app.values.yml.gotmpl <-- local overrides only
+    │    └── app.values.yml.gotmpl <-- local overrides only, plus image tag if applicable
     ├── production
     │    └── app.values.yml.gotmpl <-- base case, containing a full configuration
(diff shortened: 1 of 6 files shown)


doc/adr/0006-helmfile-base-values.md
```diff
@@ -0,0 +1,83 @@
...
+└── env
+    ├── local
+    │    └── app.values.yml.gotmpl <-- local overrides only
+    ├── production
+    │    └── app.values.yml.gotmpl <-- base case, containing a full configuration
+    └── staging
+        └── app.values.yml.gotmpl <-- staging overrides only
+```
+
+Example values for a hyptothetical app could look like this:
```
tarrow, line 49:
```suggestion
Example values for a hypothetical app could look like this:
```

doc/adr/0006-helmfile-base-values.md
```diff
@@ -0,0 +1,83 @@
...
+This is why we've been looking for ways to DRY these value files up.
+
+While the [most straight forward way][helmfile-docs] of de-duplicating values is easy to do, it creates one-off hierarchy patterns for each helmfile release, and every change to be done requires understanding yet another ad-hoc solution.
+
+
+Instead, we'd like to end up in a situation where every helmfile release uses the same pattern, making it obvious how and where to apply pending changes.
+
+## Decision
+
+Instead of defining [a least common set of values for a service in a `base` case][helmfile-docs], and then apply per-environment changes on top of this base case, __we want to always use the values for production as the base case__ and then define overrides per staging and production.
```
tarrow, line 25:
```suggestion
Instead of defining [a common set of values for a service in a `base` case][helmfile-docs], and then apply per-environment changes on top of this base case, __we want to always use the values for production as the base case__ and then define overrides per staging and production.
```

minor nitpick to perhaps clarify the meaning

m90: This was meant to say "aims at making it as minimal as possible with no regards for other aspects such as easy to understand et. al." Is there a better way to say that or do you just go "common"?

doc/adr/update-toc
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Usage: ./update-toc.sh [target=index.md] [docs_pattern=0*.md] [marker=toc]
```
tarrow, line 3: definitely a great improvement; now we just need to remember to run this script!

m90: "Dreams"

doc/adr/0006-helmfile-base-values.md
tarrow: I can't think of a nicer phrasing than common; perhaps "naively common"?

"least common" to me however implies "values that had the fewest number of things in common" which is really not what we mean.

m90: Then let's just do "common" I guess.

deer-wmde (approved):

PR #675, opened 2023-01-17
refactor(helm): derive values for staging and local from base production settings

m90:
Ticket [T326643](https://phabricator.wikimedia.org/T326643).

Applies what is described in #674 

This diffs cleanly against production, staging and local at the time of writing.


k8s/helmfile/helmfile.yaml
```diff
@@ -47,9 +47,9 @@ helmfiles:
 
 templates:
   default: &default_release
-    missingFileHandler: Warn
+    missingFileHandler: Error
```
m90, line 50: This requires an empty file to be present for environments that exactly match staging, but I guess that's to be considered a "feature"?

deer-wmde: I guess so. On the plus side, if someone wants to deviate from the prod config it's already prepared in a way (with the empty file)

PR #719, opened 2023-02-08
fix(local): elasticsearch pods need to run on the same node locally

m90:
I noticed that applying the current state to my local cluster makes ES fail to run as the affinity rules won't work for the minikube cluster:

```
Events:
  Type     Reason            Age                 From               Message
  ----     ------            ----                ----               -------
  Warning  FailedScheduling  118s                default-scheduler  0/1 nodes are available: 1 node(s) didn't match pod anti-affinity rules.
  Warning  FailedScheduling  34s (x1 over 103s)  default-scheduler  0/1 nodes are available: 1 node(s) didn't match pod anti-affinity rules.
```

This PR makes local env use `soft` affinity instead.

### DIFF
--- a/k8s/helmfile/env/staging/elasticsearch.values.yaml.gotmpl
+++ b/k8s/helmfile/env/staging/elasticsearch.values.yaml.gotmpl
@@ -2,8 +2,6 @@ imageTag: "6.8.23-wmde.6"
 
 esJavaOpts: "-Xms2g -Xmx2g"
 
-antiAffinity: hard
-
 resources:
   requests:
     cpu: "500m"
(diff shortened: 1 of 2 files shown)


k8s/helmfile/env/staging/elasticsearch.values.yaml.gotmpl
```diff
@@ -2,8 +2,6 @@ imageTag: "6.8.23-wmde.6"
 
 esJavaOpts: "-Xms2g -Xmx2g"
 
-antiAffinity: hard
```
m90, line 5: This deletion is a drive by, removing the production default of `hard` repeated in the staging values.

tarrow (approved):

PR #721, opened 2023-02-09
staging k8s: add debug logging to log4j

tarrow:


tarrow:
Idea generally taken from https://stackoverflow.com/a/60543278

m90 (approved):

PR #722, opened 2023-02-09
chore(staging-es): try different approach of arriving at debug log level

m90:
#721 somehow silenced all logging. Giving it another try as described here: https://stackoverflow.com/a/65181735/797194


tarrow (approved):

tarrow (approved):
yeah, debug better than trace for now

PR #725, opened 2023-02-09
chore(es-staging): debug does not yield any messages when stuck in booting

m90:
Even after #724, there still 8 minutes of total silence, so let's see if TRACE levels will show us anything here.


m90:
Maybe this _is_ too noisy.

PR #727, opened 2023-02-09
chore(es-staging): remove all failed attempts at debugging boot slowness

m90:
Seems none of this made any real-world impact, so we can run production settings in the staging env again.


tarrow (approved):
Optimism lost is always a state state of affairs

PR #737, opened 2023-02-16
Set up local Prometheus, add exporter for ElasticSearch

m90:


k8s/helmfile/prometheus.yaml
```diff
@@ -0,0 +1,18 @@
+releases:
+  - name: kube-prometheus-stack
+    namespace: monitoring
```
m90, line 3: Putting this in a different namespace is stolen from #446 - do we want it?

tarrow: I think yes we do want this, we'd also want to put the prometheus exporters in this namespace as well as the Pod/Service monitoring defintions

k8s/helmfile/helmfile.yaml
```diff
@@ -198,9 +203,22 @@ releases:
...
+  - name: prometheus-elasticsearch-exporter
+    namespace: default
+    chart: prometheus-community/prometheus-elasticsearch-exporter
+    version: 5.0.0
+    <<: *default_release
+
   ################################
   # LOCAL ONLY
   ################################
+  - name: kube-prometheus-stack
```
tarrow, line 215: This *probably* wants a version pin (amazingly there was an update even today).

However one thing we are very bad is is then rolling forwards with these version updates

k8s/helmfile/prometheus-operator.yaml
```diff
@@ -0,0 +1,4 @@
+releases:
+  - name: prometheus-operator-crds
```
tarrow, line 2: similarly, this probably wants a version pin

k8s/helmfile/helmfile.yaml
```diff
@@ -198,9 +203,23 @@ releases:
     chart: ./../../charts/stats-cron
     <<: *default_release
 
+  - name: prometheus-elasticsearch-exporter
```
tarrow, line 206: This also wants to only be installed in the local env for now; at the moment this also adds the exporter to all environments

m90: Isn't that what we want?

tarrow: We were historically favouring PRs touching the smallest number of environments at a time.

The reason being that if, or example, this fails to apply to production (e.g for some complicated reason due to CRDs [which I think it actually will because the prometheus-operator.yaml *isn't* being applied to production]) then we end up reverting this whole PR (or blocking other changes being made to cluster)

Of course "in the fullness of time" (probably very soon) we do want this in both staging and production

k8s/helmfile/prometheus-operator.yaml
```diff
@@ -0,0 +1,6 @@
...
+    namespace: monitoring
```
tarrow, line 3: it's probably fine and desired but trying to apply this to a cluster that already had an earlier version of this PR on it results in:
```
Error: Failed to render chart: exit status 1: Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition "alertmanagerconfigs.monitoring.coreos.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "monitoring": current value is "default"
  Error: plugin "diff" exited with error
```

m90: So what I did just this morning was:

- delete local cluster
- create new cluster from scratch
- apply main
- checkout this branch
- apply this branch

which passed for me without any issues. Not sure why yours is now complaining about duplicate CRDs?

tarrow (approved):
Lets go ahead with this

PR #761, opened 2023-02-27
fix(monitoring): es exporter crashes when running with limited resources

m90:
Since #759, the ElasticSearch exporter is stuck in CrashLoopBackoff. While the logs do not show anything suspicious, I assume this is caused by resource limits.


k8s/helmfile/env/production/prometheus-elasticsearch-exporter.values.yaml.gotmpl
```diff
@@ -4,11 +4,11 @@ es:
 
 resources:
   requests:
-    cpu: 100m
-    memory: 256Mi
+    cpu: 250m
```
m90, line 7: This is mostly just guessing.

deer-wmde (approved):

PR #765, opened 2023-03-01
chore(local): prometheus defaults exceed available resources

m90:
Sometimes, I still have the prometheus-operator failing to create the prometheus pods in local development. There is no real indication as of why, but lowering resource limits seems to help getting it up and running consistently.


deer-wmde (approved):
Interesting, let's see if this fixes it.

I also just saw that it was erroring on my local cluster but it also wasn't up to date and a `make apply-local` fixed it, so I'm not sure in what state it just was

PR #766, opened 2023-03-02
docs(adr): propose adopting an istio service mesh

m90:
Ticket [T330862](https://phabricator.wikimedia.org/T330862)


rosalieper (approved):
I am happy with the choice.

PR #767, opened 2023-03-02
chore(ci): checkout v2 uses deprecated node version

m90:
Running the `test` action in this repository currently results in a warning:

image

Instead of changing the default node version in the v2 checkout action, we can also just upgrade to v3.


tarrow:
Seems smart; I'll also turn on dependabot for this repo since we if we can avoid having to do this manually I think we should

PR #768, opened 2023-03-06
Set up Istio service mesh and add monitoring objects

m90:
Ticket https://phabricator.wikimedia.org/T330863


charts/istio-monitors/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
```
m90, line 1: I'm not a huge fan of this chart being introduced as Istio things are now scattered across different places. Is there an easy way to extend the Istio chart and pass these objects there instead? The istio chart does not expose such an interface unfortunately, so we'd need something that's built into Helm itself instead.

tarrow: My understanding but we didn't use this pattern before is that a "subchart" type re-use of the off the shelf chart in combination with our own parent (or possibly our own parent and our own sub "monitoring" chart) would be what most people would do. See: https://helm.sh/docs/chart_template_guide/subcharts_and_globals/

m90: Let me try getting that to work.

charts/istio-service-mesh/templates/pod-monitor.yaml
```diff
@@ -0,0 +1,39 @@
+{{ if .Values.podMonitor.enabled }}
```
m90, line 1: These specs are taken from https://github.com/istio/istio/tree/master/samples/addons/extras with a little configuration from our end sprinkled on top.

k8s/helmfile/env/local/elasticsearch.values.yaml.gotmpl
```diff
@@ -1,3 +1,6 @@
+labels:
+  sidecar.istio.io/inject: "true"
```
m90, line 2: This means that when testing this locally, you will need to delete the `elasticsearch-master-0|1|2` pods manually in order to trigger the proxy injection. Alternatively, you can remove the label and re-apply, which will then trigger proxy creation as the webhook is listening.

k8s/helmfile/env/local/istio-service-mesh.values.yaml.gotmpl
```diff
@@ -0,0 +1,19 @@
...
+  namespace: monitoring
+
+serviceMonitor:
+  namespace: monitoring
+
+istiod:
+  pilot:
+    resources:
+      requests:
+        cpu: 125m
```
tarrow, line 11: I tried quickly to answer this myself but figured asking would be better:
Where did these numbers come from? Do you think they are right after some experimenting or are these some defaults from somewhere?

m90: Those are mostly educated guesses from what I've seen on my machine, however I obviously couldn't do this for staging or production.

The reason I did changes there is that the values seemed _very_ high to me (when comparing to what I've seen sth like nginx consume). They might be totally wrong though.

k8s/helmfile/helmfile.yaml
```diff
@@ -217,6 +222,12 @@ releases:
     version: 5.0.0
     <<: *default_release
 
+  - name: istio-service-mesh
+    namespace: istio-system
+    installed: {{ ne .Environment.Name "production" | toYaml }}
+    chart: ./../../charts/istio-service-mesh
```
tarrow, line 228: This is totally fine but we should be slightly aware of hassle of putting things here; it can make rolling updates in the future hard since `main` of this repo is the truth we build all environments from; we should be aware that if we end up wanting a new version of this chart in one environment but not the others we'll need to move it to the charts repo (or duplicate it) etc.

m90: Good point, didn't think of this. There probably is no way to pass values to `Chart.yaml` itself?

PR #769, opened 2023-03-06
Add Github Actions Dependabot Config

tarrow:


m90 (approved):
Seems like a good idea,

PR #780, opened 2023-03-13
feat(monitoring): collect service metrics for ui

m90:
Requires #779 and https://github.com/wbstack/charts/pull/112

UI can serve as a better canary than ElasticSearch itself, allowing us to roll back changes more easily in case something unexpected goes wrong.


m90:
Seems to work as expected when deployed locally.

tarrow (approved):

PR #782, opened 2023-03-13
(makefile) Add `helmfile-sync`

deer-wmde:
Adds `helmfile-sync` make target and some documentation.

Sometimes an initial `make apply-local` times out (or something) and helm seems to be confused on what is actually deployed and what is not. An usual symptom of this is that the API pods aren't deployed.


deer-wmde:
Should this target maybe even run after an initial `make apply-local`? Maybe we could define a `setup-local` or something?

m90 (approved):

PR #785, opened 2023-03-14
chore(elasticsearch): try raising limit for fetch_shard_store threadpool in staging

m90:
Ticket: https://phabricator.wikimedia.org/T328740

It seems the `fetch_shard_store` thread pools max out their limit (currently 20 in production) when a sibling node is currently booting / recovering. Let's see if raising this limit can help us make the startup procedure a bit faster.


tarrow (approved):
Let's try it at least

PR #793, opened 2023-03-21
Update api to use 1.38 MediaWiki

m90:
~~__Important__: Before merging and deploying this, all wikis should be put into read-only mode.~~

Requires #794 

Ticket: https://phabricator.wikimedia.org/T324575

This deploys a 1.38 MediaWiki in parallel to the old 1.38 version. Also, api is updated to use a new image that is using the 1.38 database schema, and also creates new wikis using 1.38.

All existing wikis will still be routed to the 1.37 MediaWiki as they still require manual updates to happen.


deer-wmde (approved):
fwiw this cleanly applies locally (I experimentally copy pasted the changed production files into the local files instead)

PR #795, opened 2023-03-22
chore(mediawiki): uninstall mediawiki 1.37 release

m90:
To be done once the production migration to 1.38 is finished.


deer-wmde:
also bonus points for keeping in mind that just removing it from helmfile isn't sufficient :D :100:

PR #813, opened 2023-03-23
Add scripts for upgrading wikis

deer-wmde:
note: as of right now these scripts weren't battle tested yet

### DIFF
--- a/bin/wiki-readonly
+++ b/bin/wiki-readonly
@@ -21,7 +21,6 @@ Help() {
     echo "example 1: 'cat domains.txt | ./wiki-readonly'"
     echo "example 2: 'echo foobar.wikibase.cloud | ./wiki-readonly'"
     echo
-
 }
 
 while getopts "d:v:u" option; do 
--- a/bin/wiki-upgrade
+++ b/bin/wiki-upgrade
@@ -88,7 +88,7 @@ for WIKI_DOMAIN in $WIKI_DOMAINS; do
 done
 
 echo
-echo -n "Waiting 10 seconds ... "
+echo "Waiting 10 seconds ..."
 echo
 
 sleep 10


bin/lock-wikis
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# This script can be used to set the read-only setting for a list of wikis.
+# 
+# `DOMAIN_FILE_LIST` is just the path to a plain-text file containing a list of wikibase.cloud domains, each line containing 1 wiki domain
```
m90, line 5: Suggestion: Instead of forcing users to provide a file, we could also use pipes, allowing users to use any mechanism they want, e.g.:

```
cat domains.txt | lock-wikis
echo "testthis.wikibase.cloud" | lock-wikis
kubectl exec "whatever" | lock-wikis
./some-command-from-the-future | lock-wikis
```

bin/lock-wikis
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
m90, line 1: Question: Is there a reason to explicitly use `bash` when no bashisms are used in the scripts? I.e. would `/bin/sh` also do and allow for greater portability?

bin/upgrade-wikis
```diff
@@ -0,0 +1,26 @@
...
+# ```
+# foo.wikibase.cloud 
+# bar.wikibase.cloud
+# ``
+
+set -exu
+
+# These need to be adjusted for each upgrade
+DOMAIN_FILE_LIST='./domains.txt'
+DB_VERSION_OLD='mw1.37-fp-wbs1'
```
m90, line 19: Suggestion: should these be passed as options, e.g.:

```
cat domains.txt | ./upgrade-wikis -f mw1.37-fp-wbs1 -t mw1.38-wbs1 -b mediawiki-138
```

deer-wmde: I'd love that but I always struggle with implementing something like this cleanly and readable in bash

bin/lock-wikis
deer-wmde: Makes sense to me. Problem is that I always have been using bash for scripting so I don't know what is missing with sh :D

bin/lock-wikis
deer-wmde: yes I even started writing it like that but I just used $* to accept the list - and IIRC theres *some* limit to number of arguments or something? I just went for boring but safe stuff here (according to my limited bash scripting skills)

m90: > yes I even started writing it like that but I just used $* to accept the list 

I think `read` should give us piping in unlimited ways:

```sh
while read line; do
  echo "i'm a parrot: $line`
done
```

bin/upgrade-wikis
m90: I guess `getopts` is the way to go (as seen in other scripts here as well), which coincidentally means we _do need_ bash to run this:

```
DB_VERSION_TO='mw1.37-fp-wbs1'
DB_VERSION_FROM='mw1.38-wbs1'
MW_BACKEND='mediawiki-138'

# :x means there is value associated to the flag, x only would be a boolean flag
while getopts ":f:t:b" option; do 
   case $option in
      f)
         DB_VERSION_FROM=$OPTARG;;
      t)
         DB_VERSION_TO=$OPTARG;;
      b)
         MW_BACKEND=$OPTARG;;
      # optionally add a default case that exits 1 on unknown flags?
   esac
done
```

bin/lock-wikis
m90: The one major difference I have encountered is that bash knows `[[` when sh only knows `[`. Also `sh` doesn't know about `set -o pipefail` as well as `getopts` (see below).

deer-wmde:
I tried to refactor the scripts. I might have actually made it worse because now they are way more complex. I tried to think of what would benefit us in the situation of actually running these during upgrading lots of wikis, so I:
- added some `date` lines
- keep setting `set -x` only for calls to kubectl
- print out domains and used variables for upgrading (so you can review them)

some opinions on how it is now:
- getopts - I don't like short option names and I couldn't find out if/how it is possible to use long names with getopts
- it seems using `while read` as it is now, the domains **have to** come in separated by a newline  
- the review setup at the beginning is basically useless with a timeout, but we also can't use another `read` without breaking the piping functionality... ugh

m90:
> it seems using while read as it is now, the domains have to come in separated by a newline

I would argue that's ok, but if really wanted to, you could check if fd0 (stdin) is open or closed and then proceed according to that check:

```
$ [ -t 0 ] && echo "nope"                                                                                                                                              production
nope
$ echo "stdin" | [ -t 0 ] && echo "nope"  
```

> I don't like short option names and I couldn't find out if/how it is possible to use long names with getopts

No long option names with getopts I'm afraid. "Use Python" is probably not the most helpful advice either.

Personally, I think it's ok for three options and still nicer than env vars.

deer-wmde:
> I would argue that's ok, but if really wanted to, you could check if fd0 (stdin) is open or closed and then proceed according to that check:

Nice, I didn't know about that! So we could exit if nothing gets piped into it, right? I think I'd like that as some kind of guardrail

m90:
> So we could exit if nothing gets piped into it, right?

We could either exit or treat `$1` as a single domain. Exiting is probably easier.

deer-wmde:
> > So we could exit if nothing gets piped into it, right?
> 
> We could either exit or treat `$1` as a single domain. Exiting is probably easier.

ahh! I like that!

tarrow:
Particularly if we already used and benefited from these scripts I'm keen to seem them make their way into VCS so please don't see this as a blocker to merging

In general I wonder, however, if we should ask ourselves if this functionality would be better suited to living in the platform API?

I could imagine us adding a job there that sets a wiki or wikis reads only. Similarly I could imagine us adjusting the existing update job to take a list.

I think the real blocker to this work being mostly kept in the platform API is the inability to adjust the backend routing. I believe this was something Adam was talking about in the past doing by having nginx calling the platform api to determine this routing.

PR #822, opened 2023-03-29
k8s: remove istio from elasticsearch

tarrow:
In order to test the hypothesis that istio
is causing some errors in pinging between pods
lets try removing it in production and staging


deer-wmde (approved):
does this require to redeploy the ES nodes?

tarrow:
what do you mean by redeploy? I think this will mean a new release and new pods.

PR #826, opened 2023-03-30
Add job to rebuild ES from specific point in time

tarrow:


tarrow:
Draft since not yet managed to test this thing locally

m90 (approved):
I think it'd be nice to somehow document that `WBS_FORCE_SEARCH_INDEX_FROM` is expected to be a date in the form of `Start date of reindex in YYYY-mm-ddTHH:mm:ssZ`.

Other than that, no complaints.

tarrow:
I'll add a comment to the file since I'm nut sure I can think of the best way to quickly validate this.

In any case in the event you use an unacceptable format the error will read like:
```
Wikimedia\Timestamp\TimestampException from line 240 of /var/www/html/w/vendor/wikimedia/timestamp/src/ConvertibleTimestamp.php: Wikimedia\Timestamp\ConvertibleTimestamp::setTimestamp: Invalid timestamp - 2023-03-30T11:45:46Zcat
#0 /var/www/html/w/vendor/wikimedia/timestamp/src/ConvertibleTimestamp.php(165): Wikimedia\Timestamp\ConvertibleTimestamp->setTimestamp('2023-03-30T11:4...')
#1 /var/www/html/w/extensions/CirrusSearch/maintenance/ForceSearchIndex.php(158): Wikimedia\Timestamp\ConvertibleTimestamp->__construct('2023-03-30T11:4...')
#2 /var/www/html/w/maintenance/doMaintenance.php(118): CirrusSearch\Maintenance\ForceSearchIndex->execute()
#3 /var/www/html/w/extensions/CirrusSearch/maintenance/ForceSearchIndex.php(686): require_once('/var/www/html/w...')
#4 {main}
```

The acceptable formats are actually more vague than specified in the maint script docs though since internally (as you see in the stack trace) it uses ConvertibleTimestamp. This means for example non-UTC timestamps can be used

m90:
I stole my assumptions about the format from here: https://github.com/wikimedia/mediawiki-extensions-CirrusSearch/blob/fc438bf13255981f87a14b862ef0dd68647b4604/maintenance/ForceSearchIndex.php#L83 - is that not the correct place to look?

tarrow:
that is indeed where to look; however the comment there is "actually" wrong. i.e. dates other than that format will also work

PR #828, opened 2023-03-30
GHAction: Explicit version of mattermost action

tarrow:


deer-wmde:
makes sense to me. I hope that dependabot picks it up though :D

deer-wmde (approved):

PR #830, opened 2023-04-03
fix(elasticsearch): port exclusion rules need to be given in annotation

m90:
This was missed in #824 

See: https://istio.io/latest/docs/reference/config/annotations/


tarrow (approved):
Lets try the thing

PR #833, opened 2023-04-03
fix(elasticsearch): try using rules from es docs in staging

m90:
This would be an alternative to #832 as per the docs here: https://www.elastic.co/guide/en/cloud-on-k8s/2.2/k8s-service-mesh-istio.html


deer-wmde (approved):
lets see what happens :)

PR #835, opened 2023-04-03
Hard disable Istio mTLS in staging

m90:
Ticket https://phabricator.wikimedia.org/T333807

The default value of `PERMISSIVE` should work as per documentation (as it allows for both encrypted as well as unencrypted traffic), however this is not what we are seeing in staging. Instead, we can try hard disabling mTLS and see if it makes a difference.

### DIFF
--- a/charts/istio-service-mesh/templates/peer-authentication.yaml
+++ b/charts/istio-service-mesh/templates/peer-authentication.yaml
@@ -1,10 +1,8 @@
 {{ if .Values.peerAuthentication.enabled }}
+{{- range .Values.peerAuthentication.objects }}
 apiVersion: security.istio.io/v1beta1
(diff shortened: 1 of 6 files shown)


tarrow:
I see no reason not to try this just as it is on staging but I have to admit it's not clear to me exactly what's happening here or how this patch is disabling mTLS.

If I understand correctly from my reading of https://istio.io/latest/docs/reference/config/security/peer_authentication/  before this is merged we have no `PeerAuthentication` objects at all (I confirmed with `kubectl`) but somehow this still leaves us defaulting to mTLS being enabled in the permissive mode. This is confusing to me because the first example is titled
> Policy to allow mTLS traffic for all workloads under namespace foo:

Which makes me think that mTLS would not be allowed by default.

PR #837, opened 2023-04-04
docs: Add ElasticHQ to tools.md

deer-wmde:
Adds some documentation for using http://www.elastichq.org/


m90 (approved):
Thanks for adding :1st_place_medal:

PR #842, opened 2023-04-12
local: update skaffold mw conf to 1.39

deer-wmde:
The local skaffold conf still uses mw 1.38 even though we already use 1.39 locally as the main mw release.

Also I renamed the Makefile target so that it is obvious which mw version is gonna get deployed (I got bitten by this several times now :P)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,8 +96,8 @@ test: # @HELP Run yamllint tests against the repository
 test:
 	yamllint --no-warnings .
 
-skaffold-mediawiki-139: skaffold-mediawiki-139
-skaffold-mediawiki-139: # @HELP Deploy the local mediawiki image using skaffold
+skaffold-mediawiki-138: # @HELP Deploy the local mediawiki 1.38 image using skaffold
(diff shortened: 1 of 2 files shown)


tarrow (changes requested):
Generally seems like a smart idea to me (I also had something similar hacked in locally) but I was thinking it could make sense to add this in addition to the 138 one? Just given that we don't actually run 1.39 on production yet.

It might be nice to have it still there if we urgently have to unpick a bug in 1.38

deer-wmde:
> Generally seems like a smart idea to me (I also had something similar hacked in locally) but I was thinking it could make sense to add this in addition to the 138 one? Just given that we don't actually run 1.39 on production yet.
> 
> It might be nice to have it still there if we urgently have to unpick a bug in 1.38

Added it. Note however that neither target is guaranteeing to deploy the right version to the right deployment (as the actual mediawiki version in the image that skaffold builds, depends on what is checked out in your local mediawiki repo)

tarrow (approved):

tarrow:
hehe, yeah. It will build whatever is there and shim it into every pod in that release

PR #848, opened 2023-04-17
feat(mediawiki): deploy 1.39 release to production

AndrewKostka:


tarrow (approved):

k8s/helmfile/env/production/platform-nginx.nginx.conf
```diff
@@ -7,7 +7,327 @@ map_hash_bucket_size 128;
 #WHERE deleted_at IS NULL
 #LIMIT 900
 map $host $mwversion {
-        default "138-app";
+        default "139-app";
```
addshore, line 10: If this list is created at the same moment as the `wikiDbUseVersion` change, then there is a possible race condition, and you may end up with a wiki created at that moment that will be running 1.38 with a 1.38db, that then starts running 1.39 code right away and will likley also be missed as part of the currently manual upgrade process.
Not to mention that this list may get out of date while the PR is awaiting code review, pushing or deployment etc

k8s/helmfile/env/production/tool-quickstatements.values.yaml.gotmpl
```diff
@@ -1,7 +1,7 @@
 replicaCount: 1
 
 platform:
-  mediawikiBackendHost: mediawiki-138-app-backend.default.svc.cluster.local
+  mediawikiBackendHost: mediawiki-139-app-backend.default.svc.cluster.local
```
addshore, line 4: As noted above, there is more of a chance of things being broken for sites if pointing to new code while old DB schemas are used.
Though the "real" fix for the oauth tools lies in having some service infront of all mediawiki services directing things to the correct versions, and doing that autmatically.

k8s/helmfile/env/production/platform-nginx.nginx.conf
tarrow: yeah, this race exists but we do explicitly check for it. We did this by confirming that this list was still the same after deploying this. Changing wikiDbUseVersion first has a similar issue I think: a new database would be being read by an old version of the code by anyone who creates a wiki around this time.

k8s/helmfile/env/production/api.values.yaml.gotmpl
```diff
@@ -10,7 +10,7 @@ replicaCount:
 useProbes: true
 
 platform:
-  backendMwHost: mediawiki-138-app-backend.default.svc.cluster.local
+  backendMwHost: mediawiki-139-app-backend.default.svc.cluster.local
```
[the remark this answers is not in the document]
addshore: I think things can still break in this situation yes, but I'd expect them to be less likely.
But it really depends on what is happening during the update.
As written in another comment below, really infront of all mediawiki services, (including quickstatements etc) there should be switching based on mw version so that everything always aligns on the version.

k8s/helmfile/env/production/platform-nginx.nginx.conf
addshore: Indeed, though as written in other places, its highly probable that old code works with new schema, and quite probably that the other way around things will be broken.
So the steps I have taken in the past is probably something like
 - Migrate a few wikis to set out how migration is going to work, and explicitly set these to use new service in the nginx config
  - Once confident, switch default wiki creation to the new version
  - Once deployed query for all sites that are marked as using the old DB versions adding them to the nginx for the old version and setting the default to the new version.
  
  Then no double check is required and it's only in the situation of old code new schema for primary traffic for sites.

PR #875, opened 2023-04-19
Route MediaWiki traffic by querying version from Platform API

AndrewKostka:
Depends on https://github.com/wbstack/api/pull/599, https://github.com/wmde/wbaas-deploy/pull/1997, and https://github.com/wmde/wbaas-deploy/pull/1996

### DIFF
--- a/k8s/helmfile/env/local/platform-nginx.nginx.conf
+++ b/k8s/helmfile/env/local/platform-nginx.nginx.conf
@@ -96,8 +96,8 @@ server {
 
         location = /version_error {
                 internal;
-                add_header Content-Type text/plain;
-                return 400 'The requested wiki "$host" may never have existed or may now be deleted.';
+                default_type text/html;
+                return 404 '<body>You have requested the domain: $host. But that wiki can not currently be loaded.<br/>It may never have existed or it might now be deleted.<br/>You can check the platform status at <a href="https://status.wikibase.cloud">status.wikibase.cloud</a>.</body>';
         }
 
         location = /robots.txt {
--- a/k8s/helmfile/env/production/platform-nginx.nginx.conf
+++ b/k8s/helmfile/env/production/platform-nginx.nginx.conf
@@ -257,8 +257,8 @@ server {
 
         location = /version_error {
                 internal;
-                add_header Content-Type text/plain;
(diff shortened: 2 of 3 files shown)


k8s/helmfile/env/local/platform-nginx.nginx.conf
```diff
@@ -153,9 +164,11 @@ server {
         }
 
         location ~* "^/()(.*)" {
+                auth_request /version;
```
tarrow, line 167: statement: I think we talked about this way back when but I couldn't find a link to it. An alternative to making an auth request each time might be to use the nginx lua setup. This usecase is one of the ones they acknowledge in it: https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#typical-uses:~:text=fetching%20backend%20information%20from%20external%20storage%20backends%20(like%20redis%2C%20memcached%2C%20mysql%2C%20postgresql)%20and%20use%20that%20information%20to%20choose%20which%20upstream%20backend%20to%20access%20on%2Dthe%2Dfly%2C

no drama if you don't want to change this though; it might be something we need to do if we find that the version endpoint is being hit too regularly

tarrow: smart! That's a good idea

tarrow (approved):
Not manually tested it locally; lgtm though

tarrow (approved):

k8s/helmfile/env/production/platform-nginx.nginx.conf
```diff
@@ -243,6 +241,26 @@ server {
...
+                proxy_cache_key $host;
+                proxy_cache_valid any 5m;
+                proxy_ignore_headers Cache-Control;
+                add_header X-Cache-Status $upstream_cache_status;
+	}
+
+        location = /version_error {
+                internal;
+                default_type text/html;
+                return 404 '<body>You have requested the domain: $host. But that wiki can not currently be loaded.<br/>It may never have existed or it might now be deleted.<br/>You can check the platform status at <a href="https://status.wikibase.cloud">status.wikibase.cloud</a>.</body>';
```
tarrow, line 261: the current error is also including `Failure looking up wiki asdfasd.wikibase.cloud.` (for example) but I don't think tweaking this a little bit matters

PR #876, opened 2023-04-20
feat(alarms): add tf module for creating alarms based on prometheus metrics

m90:
Ticket https://phabricator.wikimedia.org/T334184


m90:
After reading https://cloud.google.com/monitoring/alerts/concepts-indepth it seems we will have to live with two incidents being created:

image

m90:
There's also an issue about this in GCP support https://issuetracker.google.com/issues/194210681?pli=1

PR #878, opened 2023-04-24
chore(tf): merge monitoring and metric-alarms modules

m90:
#876 needlessly introduced a new module when we already have `monitoring`. This PR merges the content of the two.


tf/modules/monitoring/variables.tf
```diff
@@ -1,36 +1,41 @@
 variable "cluster_name" {
-  type = string
+  type        = string
```
m90, line 2: This is my editor running `terraform fmt` on the source files, which I guess is ok.

deer-wmde (approved):

PR #885, opened 2023-05-10
feat(k8s): add namespace for api jobs and add secrets to both namespaces

m90:
Requires: https://github.com/wmde/wbaas-deploy/pull/892

Ticket: https://phabricator.wikimedia.org/T330389

As discussed, we want to create a dedicated namespace that contains the rather noisy job objects. As there seems to be no official way of sharing secrets across namespaces, I went for defining the namespace in terraform and also duplicating the secrets across namespaces.

~~Leaving this a draft as it contains dev junk still.~~

### DIFF
--- a/tf/env/local/namespaces.tf
+++ b/tf/env/local/namespaces.tf
@@ -3,3 +3,16 @@ resource "kubernetes_namespace" "api-job-namespace" {
     name = "api-jobs"
   }
 }
+
+resource "kubernetes_resource_quota" "api-jobs-podquota" {
+  metadata {
+    name      = "api-jobs-podquota"
+    namespace = kubernetes_namespace.api-job-namespace.metadata[0].name
+  }
+  spec {
+    hard = {
+      pods = 2
+    }
+    scopes = ["BestEffort"]
+  }
+}
--- a/tf/env/local/namespaces.tf
+++ b/tf/env/local/namespaces.tf
@@ -1,0 +1,5 @@
+resource "kubernetes_namespace" "api-job-namespace" {
+  metadata {
+    name = "api-jobs"
+  }
+}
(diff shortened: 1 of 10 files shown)


tf/env/local/namespaces.tf
```diff
@@ -0,0 +1,5 @@
+resource "kubernetes_namespace" "api-job-namespace" {
+  metadata {
+    name = "api-jobs"
+  }
+}
```
deer-wmde, line 5: why is this needed for local / not needed for staging & prod?

m90: It should also go into staging and prod, I just did not finish this PR yet :)

deer-wmde: ahh whoops sorry :sweat_smile:

tf/env/local/secrets-api.tf
```diff
@@ -29,13 +29,14 @@ resource "random_password" "api-app-jwt-secret" {
 }
 
 resource "kubernetes_secret" "api-app-secrets" {
+  for_each = toset(["default", "api-jobs"])
   metadata {
-    name = "api-app-secrets"
-    namespace = "default"
+    name      = "api-app-secrets"
+    namespace = each.value
```
m90, line 35: While `terraform` can `plan` this change properly, it seems dropping and recreating a secret like this _sometimes_ creates race conditions in Kubernetes, making tf fail at apply-time. Not sure what to do about it right now to be honest.

m90: This can be fixed using `moved` blocks.

deer-wmde (approved):
Can confirm this worked locally for me! details: https://phabricator.wikimedia.org/T330389#8895083

deer-wmde (approved):
Can confirm this worked locally for me! details: https://phabricator.wikimedia.org/T330389#8895083

PR #886, opened 2023-05-11
Add local env convenience hacks

deer-wmde:
Some days ago I told you about these scripts I created for convenience to:
- recreate the local cluster without interaction
- create a local user account which is rady to go

Hope this helps.

### DIFF
--- a/bin/local/create-local-user.sh
+++ b/bin/local/create-local-user.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+set -e
+
+### failsafe logic to exit in case we are not running in our local minikube context
+KUBE_CONTEXT=$(kubectl config current-context)
+
+echo "Current kube context: '${KUBE_CONTEXT}'"
+
+if [[ "${KUBE_CONTEXT}" != "minikube-wbaas" ]]; then
+    echo "Error: wrong kube context. Use this script only within 'minikube-wbaas'!"
+    exit 1
+fi
+#####################################################################################
+
+# This script should create and verify a user account on a local wbaas cluster with the credentials below. Do not use this in production! 
+
+USER_CODE="create-local-user"
+USER_MAIL="jane.doe@wikimedia.de"
+USER_PASS="wikiwikiwiki"
+
+TEST_RESPONSE=$(curl -s http://api.wbaas.localhost/healthz)
+if [[ "${TEST_RESPONSE}" != "It's Alive" ]]; then
+    echo "Error: local api is not available. Is the minikube tunnel open? ('make minikube-tunnel')"
+    exit 2
+fi
+
+echo "> Creating local invite code '${USER_CODE}' ..."
+kubectl --context ${KUBE_CONTEXT} exec deployments/api-queue -- php artisan wbs-invitation:create ${USER_CODE}
+
+echo "> Registering local user '${USER_MAIL}' ..."
+curl "http://api.wbaas.localhost/user/register" \ 
+    -X POST \ 
+    -H "Content-Type: application/json" \ 
+    --data-raw "{"email":"${USER_MAIL}","password":"${USER_PASS}","invite":"${USER_CODE}","recaptcha":"localdebug"}"
+
+echo "> Verifying local user '${USER_MAIL}' ..."
+kubectl --context ${KUBE_CONTEXT} exec deployments/api-app-backend -- \ 
+    php artisan tinker --execute \ 
+        "User::firstWhere(["email" => "${USER_MAIL}"])->update(["verified" => 1])"'
--- a/bin/local/nuke-local.sh
+++ b/bin/local/nuke-local.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+set -e
+
+### failsafe logic to exit in case we are not running in our local minikube context
+KUBE_CONTEXT=$(kubectl config current-context)
+
+echo "Current kube context: '${KUBE_CONTEXT}'"
+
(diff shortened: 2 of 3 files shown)


bin/local/create-local-user.sh
```diff
@@ -0,0 +1,40 @@
...
+
+if [[ "${KUBE_CONTEXT}" != "minikube-wbaas" ]]; then
+    echo "Error: wrong kube context. Use this script only within 'minikube-wbaas'!"
+    exit 1
+fi
+#####################################################################################
+
+# This script should create and verify a user account on a local wbaas cluster with the credentials below. Do not use this in production! 
+
+USER_CODE="create-local-user"
```
m90, line 18: Would it be a thing to make these configurable?

```sh
USER_CODE=${USER_CODE:-create-local-user}
```

bin/local/create-local-user.sh
```diff
@@ -0,0 +1,40 @@
...
+USER_MAIL="jane.doe@wikimedia.de"
+USER_PASS="wikiwikiwiki"
+
+TEST_RESPONSE=$(curl -s http://api.wbaas.localhost/healthz)
```
m90, line 22: Could we somehow make this work without someone having to `minikube-tunnel` beforehand? I.e. ask Kubernetes to portforward to api and then do the pinging?

bin/local/nuke-local.sh
```diff
@@ -0,0 +1,36 @@
...
+
+echo "⚠️ ☠️  NUKING your local cluster in 10 seconds ... all data will be lost! (press ctrl+c NOW to cancel) ☠️ ⚠️"
+sleep 10
+
+make minikube-delete
+make minikube-start
+
+cd tf/env/local/ && terraform apply -auto-approve # bad practice - you won't have a second chance to check for destructive actions
+
+cd ../../../k8s/helmfile/
```
m90, line 26: This means you can't call this script from any other working directory but `bin/local` if I'm not mistaken.

Maybe we can use sth like

```sh
cd $(dirname $0)/../../../k8s/helmfile/
```

instead so that it can be called from anywhere?

bin/local/create-local-user.sh
```diff
@@ -0,0 +1,40 @@
...
+echo "> Registering local user '${USER_MAIL}' ..."
+curl "http://api.wbaas.localhost/user/register" \ 
+    -X POST \ 
+    -H "Content-Type: application/json" \ 
+    --data-raw "{"email":"${USER_MAIL}","password":"${USER_PASS}","invite":"${USER_CODE}","recaptcha":"localdebug"}"
+
+echo "> Verifying local user '${USER_MAIL}' ..."
+kubectl --context ${KUBE_CONTEXT} exec deployments/api-app-backend -- \ 
+    php artisan tinker --execute \ 
+        "User::firstWhere(["email" => "${USER_MAIL}"])->update(["verified" => 1])"'
```
m90, line 40: So many hours of work saved :heart:

bin/local/nuke-local.sh
deer-wmde: the idea is to run these scripts from `wbaas-deploy/` root but this still makes sense to me, i'll test and add it

bin/local/create-local-user.sh
```diff
@@ -0,0 +1,43 @@
...
+    echo "Error: wrong kube context. Use this script only within 'minikube-wbaas'!"
+    exit 1
+fi
+#####################################################################################
+
+# This script should create and verify a user account on a local wbaas cluster with the credentials below. Do not use this in production! 
+
+USER_CODE="${USER_CODE:-create-local-user}"
+USER_MAIL="${USER_MAIL:-jane.doe@wikimedia.de}"
+USER_PASS="${USER_PASS:-wikiwikiwiki}"
```
deer-wmde, line 20: Seems like GitGuardian is flagging the default password again and again, maybe we should remove it?

m90: Without ever having used it, I would guess it's possible to tell GitGuardian this is an ok case?

deer-wmde: I marked it as a test credential, let's see if it keeps that in mind in the future https://dashboard.gitguardian.com/workspace/306574/incidents/6671167?sort_date=true

bin/local/create-local-user.sh
m90: Yeah it could get a little tricky. Let's resort to the current state for now if you're ok with it too.

bin/local/nuke-local.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -e
+
+### failsafe logic to exit in case we are not running in our local minikube context
+KUBE_CONTEXT=$(kubectl config current-context)
```
m90, line 6: Suggestion: as this could be helpful in basically all other scripts as well, would it be a thing to factor it out into a file other scripts can source, with an API like this:

```
source ./require-env.sh
require_env "local"
# or
require_env "staging" "production"
```

m90 (approved):
Looks good to me, I left one suggestion, but this can also be done at some later point in time if you even like it.

bin/local/nuke-local.sh
deer-wmde: yeah I'm super torn apart about stuff like this. Personally I deeply love it and even used something like this at first, but keeping our codebase complexity in mind I like to think that it's easier maintainable and understandable if we keep these self-contained :thinking: I guess there's a possibility for an ADR discussion hidden here somewhere

PR #887, opened 2023-05-11
feat(elasticsearch): add an Elasticsearch 7.10.2 cluster on staging

AndrewKostka:


tarrow (approved):
looks fine to me; definitely happier using something other than `primary` for this new cluster. Using bee names is totally fine to me too and somewhat humorous; hopefully we don't get confused by this in the future; if we do I suggest moving to including the version number (or even just integers) in the name.

PR #889, opened 2023-05-11
fix(api): remove duplicate "app.redis.prefix" key

AndrewKostka:


tarrow (approved):
To be clear this is a duplicate from this exact same block (see line 71)

PR #891, opened 2023-05-15
local: use two Elasticsearch clusters

AndrewKostka:


m90:
This is now outdated.

PR #892, opened 2023-05-16
feat(tf): allow mediawiki secrets to be created in multiple namespaces

m90:
This is prep work for #885:

Consumers of the `k8s-secrets` module can now pass additional namespaces, making Terraform create duplicate secrets in these namespaces. This is needed so that https://github.com/wbstack/api/pull/603 can run Kubernetes jobs whose spec is derived from running Mediawiki pods in a non-default namespace.


tf/env/production/kubernetes-secrets.tf
```diff
@@ -3,24 +3,24 @@ module "wbaas-k8s-secrets" {
...
     kubernetes = kubernetes.wbaas-3
   }
-  smtp_username = mailgun_domain.default.smtp_login
-  smtp_password = random_password.smtp-password.result
-  google_service_account_key_api = google_service_account_key.api.private_key
-  google_service_account_key_dns = google_service_account_key.certman-dns-cloud-solver.private_key
-  sql_password_root = random_password.sql-root-password.result
-  sql_password_replication = random_password.sql-passwords["production-replication"].result
-  sql_password_api = random_password.sql-passwords["production-api"].result
+  smtp_username                     = mailgun_domain.default.smtp_login
```
m90, line 6: My editor runs `terraform fmt` on save. Not sure if we want this.

deer-wmde (approved):

PR #893, opened 2023-05-16
(proposal) ADR: Terraform style conventions

deer-wmde:
This PR contains a new ADR, proposing to make use of the default terraform style conventions.
Additionally, all existing terraform code has been reformatted with `terraform fmt -recursive`.

### DIFF
--- a/doc/adr/0009-terraform-style-conventions.md
+++ b/doc/adr/0009-terraform-style-conventions.md
@@ -4,7 +4,7 @@ Date: 2023-05-16
 
 ## Status
 
-proposed
+accepted
 
 ## Context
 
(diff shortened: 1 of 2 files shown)


m90 (commented):
+1.000.000 from my end.

If we wanna do it, I would think we should also have a CI job that runs `terraform fmt -check` for us, leaving it up to developers on how they want to get their changes into shape (IDE, CLI, git hooks).

tarrow:
I love this idea; let's do it. I'll just pluck off the ADR and then followup with the linting fixes

tarrow (approved):
I love this; I'll try to solicit some feedback in the daily and if no objections let's get this merged today; we can follow up with PRs for linting fixes and automation to keep them in sync

PR #904, opened 2023-05-23
chore(sql): bump resource limits, use same values for limits and requests throughout

m90:
Ticket https://phabricator.wikimedia.org/T334110


tarrow (approved):
:shipit:

PR #926, opened 2023-05-31
chore(terraform): Offboarded Adam S

manicki:


manicki:
I have no idea what I am doing. Might be the entry in `tf/env/staging/remote-state/main.tf` has been meant as some educational example. Not sure if it would be meant to kept with the deleted account any way

tarrow:
looks fine to me

deer-wmde (approved):
yeah looks good. the only other places I found are in READMEs

PR #932, opened 2023-06-05
Revert "Revert "feat(k8s): add namespace for api jobs and add secrets to both namespaces (#885)" (#930)"

deer-wmde:
This reverts commit c8416a63bf9037855c8dbc51af5514975ef81e25 in order to attempt deployment again of https://github.com/wmde/wbaas-deploy/pull/885

### DIFF
--- a/tf/env/staging/namespaces.tf
+++ b/tf/env/staging/namespaces.tf
@@ -1,0 +1,18 @@
+resource "kubernetes_namespace" "api-job-namespace" {
+  metadata {
+    name = "api-jobs"
+  }
+}
+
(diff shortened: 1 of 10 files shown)


tf/env/staging/namespaces.tf
```diff
@@ -0,0 +1,18 @@
+resource "kubernetes_namespace" "api-job-namespace" {
+  metadata {
```
tarrow, line 2: This probably needs to have the wbaas-2 provider explicitly set like https://developer.hashicorp.com/terraform/language/providers/configuration#selecting-alternate-provider-configurations

e.g. `provider = kubernetes.wbaas-2`

PR #936, opened 2023-06-07
k8s: fix api helmfile release version

tarrow:
Current ternary definition appears to not work for production. Let's try the more verbose if, else and end syntax. This should be tested by a second developer to apply cleanly to both environments before merging

### DIFF
--- a/k8s/helmfile/helmfile.yaml
+++ b/k8s/helmfile/helmfile.yaml
@@ -148,7 +148,7 @@ releases:
 
   - name: api
     chart: wbstack/api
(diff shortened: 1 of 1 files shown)


deer-wmde (approved):
For staging I get the expected ES PodDisruptionBudget error and a clean diff for prod. Therefore I assume this works as intended.

tarrow:
hrm, so the yaml linting fails. I guess because the linter doesn't realise that this is actually a template rather than a plain yaml file

dati18 (approved):

PR #939, opened 2023-06-12
fix(tf): smtp credentials are missing from additional namespaces

m90:
Ticket https://phabricator.wikimedia.org/T330389

MediaWiki jobs in the `api-jobs` namespace currently fail to start up as they fail to find a specified secret:

```
secret "smtp-credentials" not found
```


tarrow (approved):
Fantastic; I also had no idea about this `moved` syntax (https://developer.hashicorp.com/terraform/language/modules/develop/refactoring). We should have been doing this all along

deer-wmde (approved):
Ahh! I wondered where the missing link is, this looks promising :+1:

PR #944, opened 2023-06-13
feat(monitoring): enable kube-state-metrics collection

m90:
When trying to monitor the deployment of #942 I noticed we currently disable collection of all Kubernetes metrics. Enabling `kube-state-metrics` will give us metrics like job count et al: https://github.com/kubernetes/kube-state-metrics/blob/31d6e8fc265b1dadf543f2ed7da27a046c50c29f/docs/README.md


deer-wmde:
I wonder if this increases costs in any significant way? I have no idea, really, the thought just popped into my mind seeing this

m90:
It does definitely increase costs, but nothing extraordinary.

PR #948, opened 2023-06-13
chore(api): bump image version to 8x.12.2

m90:
Deploys https://github.com/wbstack/api/pull/609


tarrow (approved):

m90:
>  could also remove the local and staging overrides :)

IIRC we decided to always specify a version https://github.com/wmde/wbaas-deploy/blob/main/doc/adr/0006-helmfile-base-values.md#exception-image-tag-values

PR #950, opened 2023-06-14
tf monitoring module: fix QueryService PV alerting

tarrow:
The cross_series_reducer was wrongly set to be a sum function which resulted in the unwanted sideffect of the metric doubling for windows where two series were present.

This happens, for example, when a pod changes.

Bug: T339101


deer-wmde (approved):
I *think* I understand this behavior correctly (always struggle a bit with these), it definitely sounds plausible to me. approved

dati18 (approved):

PR #952, opened 2023-06-15
local script: Add `create-local-wiki.sh`

deer-wmde:
In order to create a whole load of wikis locally (https://phabricator.wikimedia.org/T335960) I created this script. Also improved `create-local-user.sh` because I found out about [jo](https://jpmens.net/2016/03/05/a-shell-command-to-create-json-jo/)


m90 (approved):
This is still great and should be merged.

PR #960, opened 2023-06-16
feat(elasticsearch): add Elasticsearch 7.10.2 using Bitnami

AndrewKostka:
Deploy Elasticsearch 7.10.2 in our local cluster using Bitnami


tarrow:
If we go ahead and use these bitnami charts it does seem possible we will need to build ourselves a new ES image.

Do we want to use the release-pipeline to build this thing? In the past we simply used the release pipeline image because it was "already there" and compatible with the chart we were using. We may rather have our own build process for this image.

rosalieper (approved):

tarrow:
hey! did this work for you; still debugging my weird local install but `Error: can't get a valid version for repositories elasticsearch. Try changing the version constraint in Chart.yaml` is what I get spat out

PR #964, opened 2023-06-20
chore(queryservice-gateway): update image in staging and local, use updated chart

m90:
Ticket https://phabricator.wikimedia.org/T339906

Requires: https://github.com/wbstack/queryservice-gateway/pull/163 and https://github.com/wbstack/charts/pull/118


k8s/helmfile/helmfile.yaml
```diff
@@ -172,7 +172,7 @@ releases:
   - name: queryservice-gateway
     namespace: default
     chart: wbstack/queryservice-gateway
-    version: 0.1.2
+    version: 0.1.3
```
m90, line 175: This could also happen conditionally, but adding the env var to production will be a noop, so I guess it's better to update everything instead of having to wrangle the templating once more.

tarrow: I sympathise with this position; getting the templating right (/handing this seeming edge case that Denniz and I saw about struggling to have two different charts in the prod/staging helmfile) is annoying.

I'd love us to figure out a way that this doesn't feel like a corner we want to cut. After all the "idea" of staging is that we roll new code (or images) and new config there first and have some rough certainty they are working as expected before we then move forward to production.

m90: That `ternary` helper seemed like a solid solution to the problem for me. We might want to debug why this doesn't work in some setups (I have yet to see it fail).

tarrow (approved):

PR #971, opened 2023-07-03
local: improve minikube-tunnel script

deer-wmde:
Unfortunately I can't exactly pin down under which circumstances this can happen, but in the past it happened several times for me that the minikube-tunnel script remained hanging and I couldn't ctrl+c out of it.

Recently I stumbled over a [submission on HN](https://news.ycombinator.com/item?id=36400465) about using bash exit traps, and I wondered if this change would help with the above problem.

An alternative approach could be using `EXIT INT ABRT KILL TERM`

Overall this is just a minor convenience improvement. Did anyone else notice this problem as well at all?


m90 (approved):
Let's give it a try

PR #975, opened 2023-07-06
fix(platform-summary): use align_delta instead of mean

m90:
Ticket: https://phabricator.wikimedia.org/T335961

Fixup for #970

---

We also still need to change the aligner function it seems.


tarrow (approved):
Happy that this looks like the other series entered in this dashboard.

I'm definitely not really fully understanding the difference in these reducers / aligners and if this is actually the "right" choice. At least if it's wrong it will be consistently wrong across the while dashboard though

PR #980, opened 2023-07-06
chore(elasticsearch): deploy ES 7.10.2 to staging

deer-wmde:
After applying https://github.com/wmde/wbaas-deploy/pull/978 I noticed that the staging cluster wbaas-2 had too little resources available. 

This is a follow-up PR to attempt this deployment again and tweak the resources. As of the current state, scheduling works, but the pods keep getting OOM killed. Therefore I think we have to beef up our staging nodes.


m90:
Closing in favor of #986

PR #985, opened 2023-07-13
chore(cluster): bump staging node count to 6 so it can accomodate 2 ES clusters

m90:
Ticket https://phabricator.wikimedia.org/T335853

As I assume this is what we need to in production as well, this adds 3 more nodes of the same type so we can have one ES pod per node.


tarrow (approved):
This looks like a fine way to scale up; it's clearly crazy that ES is almost all of our compute load though.

PR #1009, opened 2023-07-24
staging: update google TF provider

tarrow:


m90:
These changes still need to be reflected in the lockfile, else running `terraform init` will fail like
```
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider hashicorp/google: locked provider registry.terraform.io/hashicorp/google 4.53.1 does not match configured version constraint 4.74.0; must
│ use terraform init -upgrade to allow selection of new versions
```

m90 (approved):

tarrow:
ah damn, I didn't think about that. It also tried to bring in an update for `hashucorp/random` which I rejected  but we should probably followup with

PR #1026, opened 2023-07-31
feat(cluster): update logging variant for production

m90:
Production version of #1025


m90:
Leaving this a draft until we know it works properly in staging and have coordinated a proper timeslot for the rollout.

PR #1065, opened 2023-08-07
feat(k8s): run adhoc jobs in dedicated namespace

m90:
Ticket https://phabricator.wikimedia.org/T335891

This is prep work for running the ElasticSearch 7 init jobs. Instead of manually running these one by one in the default namespace, create a new namespace for _all_ adhoc jobs which can then handle concurrency for us.

Also, move all existing jobs to run in the new namespace.

### DIFF
--- a/k8s/jobs/changeReplicationPasswordOnSecondary.yaml
+++ b/k8s/jobs/changeReplicationPasswordOnSecondary.yaml
@@ -2,6 +2,7 @@ apiVersion: batch/v1
 kind: Job
 metadata:
   generateName: change-replication-password-on-secondary-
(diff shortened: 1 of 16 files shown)


k8s/jobs/changeReplicationPasswordOnSecondary.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: batch/v1
 kind: Job
 metadata:
   generateName: change-replication-password-on-secondary-
+  namespace: adhoc-jobs
```
m90, line 5: Defining these here forces us to also create the namespace in production already, but I guess it won't hurt either.

PR #1066, opened 2023-08-07
feat(k8s): add elasticsearch init job

m90:
Ticket https://phabricator.wikimedia.org/T335891

Requires #1065

### DIFF
--- a/k8s/jobs/elasticSearchInitJob.sh
+++ b/k8s/jobs/elasticSearchInitJob.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+# WBS_DOMAIN should be set to the domain to reindex
+# CLUSTER_NAME can be set when a single cluster should be targeted. Defaults to `all`
+# MW_WRITE_ONLY_ELASTICSEARCH_HOST can be set to override the currently used host value (might be empty)
+
+tmp_file=$(mktemp --suffix=.json)
+
+MW_POD=$(kubectl get pods --field-selector='status.phase=Running' -l app.kubernetes.io/name=mediawiki,app.kubernetes.io/component=app-backend -o jsonpath="{.items[0].metadata.name}")
+MW_POD_JSON=$(kubectl get pods $MW_POD -o=json)
+echo ${MW_POD_JSON} > $tmp_file
+
+kubectl create -f elasticSearchInitJob.yaml -o=json --dry-run=client |\
+jq -s ".[0].spec.template.spec.containers[0].image = .[1].spec.containers[0].image" - $tmp_file |\
+jq ".[0].spec.template.spec.containers[0].env = .[1].spec.containers[0].env" |\
+jq ".[0]" |\
+jq ".spec.template.spec.containers[0].env += [{\"name\": \"WBS_DOMAIN\", \"value\": \"${WBS_DOMAIN}\"}]" |\
+jq ".spec.template.spec.containers[0].env += [{\"name\": \"CLUSTER_NAME\", \"value\": \"${CLUSTER_NAME:-all}\"}]" |\
+jq ".spec.template.spec.containers[0].env += [{\"name\": \"MW_WRITE_ONLY_ELASTICSEARCH_HOST\", \"value\": \"${MW_WRITE_ONLY_ELASTICSEARCH_HOST}\"}]" |\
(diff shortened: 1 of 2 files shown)


k8s/jobs/elasticSearchInitJob.sh
```diff
@@ -0,0 +1,22 @@
...
+MW_POD_JSON=$(kubectl get pods $MW_POD -o=json)
+echo ${MW_POD_JSON} > $tmp_file
+
+kubectl create -f elasticSearchInitJob.yaml -o=json --dry-run=client |\
+jq -s ".[0].spec.template.spec.containers[0].image = .[1].spec.containers[0].image" - $tmp_file |\
+jq ".[0].spec.template.spec.containers[0].env = .[1].spec.containers[0].env" |\
+jq ".[0]" |\
+jq ".spec.template.spec.containers[0].env += [{\"name\": \"WBS_DOMAIN\", \"value\": \"${WBS_DOMAIN}\"}]" |\
+jq ".spec.template.spec.containers[0].env += [{\"name\": \"CLUSTER_NAME\", \"value\": \"${CLUSTER_NAME:-all}\"}]" |\
+jq ".spec.template.spec.containers[0].env += [{\"name\": \"MW_WRITE_ONLY_ELASTICSEARCH_HOST\", \"value\": \"${MW_WRITE_ONLY_ELASTICSEARCH_HOST}\"}]" |\
```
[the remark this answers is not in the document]
tarrow: I'm not sure I have any good ideas beyond: are we stretching the limits of readable bash? Could rewrite in python or go or something but this then brings its own hassle of execution environment/compilation etc.

tarrow (approved):

tarrow:
bah, I realise I have no idea how to remove my approve but I clicked this accidentally

k8s/jobs/elasticSearchInitJob.yaml
```diff
@@ -0,0 +1,20 @@
...
+      containers:
+        - name: elasticsearch-init-job
+          command:
+            - 'bash'
+            - '-c'
+            - |
+                MW_INSTALL_PATH=/var/www/html/w/ php \
+                  /var/www/html/w/extensions/CirrusSearch/maintenance/UpdateSearchIndexConfig.php \
+                  --cluster="${CLUSTER_NAME:-all}"
+          restartPolicy: Never
```
tarrow, line 20: I failed to apply this thing locally. Error was `Error from server (BadRequest): error when creating "STDIN": Job in version "v1" cannot be handled as a Job: strict decoding error: unknown field "spec.template.spec.containers[0].restartPolicy"`

I think restartPolicy should be on the whole pod not the container. However I wonder why we want to change the restart policy? If I understand correctly running this job twice is fine and we'd probably rather in the event of temporarily slow/timeout DB etc. this thing retries than we have to manually go and kick it off again?

```suggestion
      restartPolicy: Never
```

m90: So we'd make it `OnFailure` instead?

k8s/jobs/elasticSearchInitJob.sh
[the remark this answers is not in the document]
m90: But then we'd have a lot of other problems. I agree a mid term strategy on general purpose scripting would be nice to have.

k8s/jobs/elasticSearchInitJob.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  generateName: elasticsearch-init-job-
+  namespace: adhoc-jobs
+spec:
+  template:
+    metadata:
```
deer-wmde, line 8: suggesting some influence from https://github.com/wmde/wbaas-deploy/pull/1070 here, since that is now merged
```suggestion
        ttlSecondsAfterFinished: 604800
        metadata:
```

k8s/jobs/elasticSearchInitJob.sh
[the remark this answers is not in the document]
deer-wmde: not sure if it would help a lot here (and if we want to depend on that tool) but I used `jo` previously for local scripts
- https://github.com/wmde/wbaas-deploy/blob/main/bin/local/create-local-user.sh#L36
- https://github.com/jpmens/jo

but yeah I'm also a bit turned off by these scripts in general, see https://github.com/wmde/wbaas-deploy/pull/1081

I guess we could create Laravel Commands? (not sure if that could be an appropriate substitute)

k8s/jobs/elasticSearchInitJob.yaml
m90: Love this :heart_eyes_cat:

PR #1068, opened 2023-08-08
local/staging: add prometheus-es-exporter for 7.10 cluster

tarrow:
Copies the existing exporter configuration. This targets the main elasticsearch service which only targets
the coordinating nodes.

Also add configuration with production values so that rolling forward to production will see minimal changes

After deploying this we need to follow up on the alerting and dashboards using ES metrics since we may now see places where data from both clusters are unintentionally mixed

### DIFF
--- a/k8s/helmfile/env/production/prometheus-elasticsearch-exporter-1.values.yaml.gotmpl
+++ b/k8s/helmfile/env/production/prometheus-elasticsearch-exporter-1.values.yaml.gotmpl
@@ -1,0 +1,23 @@
+es:
+  uri: http://elasticsearch-1.default.svc.cluster.local:9200
+  timeout: 90s
+
+resources:
+  requests:
+    cpu: 250m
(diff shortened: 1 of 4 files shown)


k8s/helmfile/env/production/prometheus-elasticsearch-exporter-1.values.yaml.gotmpl
```diff
@@ -0,0 +1,23 @@
+es:
+  uri: http://elasticsearch-1.default.svc.cluster.local:9200
```
m90, line 2: So we gave up on the idea of using a headless service because we just don't know which one to monitor, correct?

m90: I wonder if we can find out where this service is pointing at?

m90 (approved):

PR #1080, opened 2023-08-09
feat(local): install minio object storage via helm chart

m90:
Ticket https://phabricator.wikimedia.org/T343112

This is prep work for moving backup transfer over to an S3 client. As of this PR, there are no consumers of the MinIO service.

## How to test/use this locally

1. Apply the changes and wait for the `minio` service to become ready
1. Install the [minio client](https://min.io/docs/minio/linux/reference/minio-mc.html)
1. Open a tunnel for the service:
    ```
    kubectl port-forward svc/minio 9000:9000
    ```
1. Set an alias for your local cluster
    ```
    mc alias set minikube "http://localhost:9000/" minio $(k get secrets/minio-credentials -o json | jq -r .data.rootPassword | base64 -d)
    ```
1. Poke around, e.g. create a new bucket and upload a file
   ```
   mc mb test-bucket
   echo "oh wow" > foo.txt
   mc cp foo.txt minikube/test-bucket
   mc cat minikube/test-bucket/foo.txt
   ```

### DIFF
--- a/k8s/helmfile/env/local/minio.values.yaml.gotmpl
+++ b/k8s/helmfile/env/local/minio.values.yaml.gotmpl
@@ -1,0 +1,19 @@
+existingSecret: minio-credentials
+
+image:
+  tag: RELEASE.2023-07-07T07-13-57Z
+
+mode: standalone
+
+replicas: 1
+
+persistence:
+  size: 4Gi
+
+buckets:
+  - name: backups
+    policy: public
+
+resources:
+  requests:
+    memory: 0.5Gi
--- a/tf/env/local/secrets-minio.tf
+++ b/tf/env/local/secrets-minio.tf
@@ -1,0 +1,17 @@
+resource "random_password" "minio-password" {
+  length = 32
+}
+
+resource "kubernetes_secret" "minio-credentials" {
+  metadata {
+    name = "minio-credentials"
+  }
+
+  binary_data = {
+    "rootPassword" = base64encode(random_password.minio-password.result)
+  }
+
+  data = {
+    "rootUser" = "minio"
+  }
+}
--- a/k8s/helmfile/helmfile.yaml
+++ b/k8s/helmfile/helmfile.yaml
@@ -18,6 +18,8 @@ repositories:
     url: https://prometheus-community.github.io/helm-charts
   - name: istio
     url: https://istio-release.storage.googleapis.com/charts
+  - name: minio
+    url: https://charts.min.io
 
 environments:
   default:
@@ -257,3 +259,11 @@ releases:
     version: 5.0.1
     values:
       - env/local/mailhog.values.yaml.gotmpl
+
+  - name: minio
+    installed: {{ eq .Environment.Name "local" | toYaml }}
+    namespace: default
+    chart: minio/minio
+    version: 5.0.13
+    values:
+      - env/local/minio.values.yaml.gotmpl


tf/env/local/secrets-minio.tf
```diff
@@ -0,0 +1,17 @@
...
+  length = 32
+}
+
+resource "kubernetes_secret" "minio-credentials" {
+  metadata {
+    name = "minio-credentials"
+  }
+
+  binary_data = {
+    "rootPassword" = base64encode(random_password.minio-password.result)
```
m90, line 11: Considering this is local-only, we could also hard code a secret key. Any opinions?

k8s/helmfile/env/local/minio.values.yaml.gotmpl
```diff
@@ -0,0 +1,19 @@
+existingSecret: minio-credentials
```
m90, line 1: Slightly twisted: this create a root user off the k8s secret of the given name.

k8s/helmfile/env/local/minio.values.yaml.gotmpl
```diff
@@ -0,0 +1,19 @@
...
+
+image:
+  tag: RELEASE.2023-07-07T07-13-57Z
+
+mode: standalone
+
+replicas: 1
+
+persistence:
+  size: 4Gi
```
m90, line 11: This seems very generous, but I also have the newest laptop around here. Should this be lower?

k8s/helmfile/env/local/minio.values.yaml.gotmpl
```diff
@@ -0,0 +1,19 @@
...
+
+buckets:
+  - name: backups
+    policy: public
+
+resources:
+  requests:
+    memory: 0.5Gi
```
m90, line 19: Same here: does this create problems for others?

m90:
It would also be possible to create an ingress for the service which exposes an administration UI, but I'm not sure this is currently needed?

m90:
This can be tested in conjunction with https://github.com/wmde/wbaas-backup/pull/24 and https://github.com/wbstack/charts/pull/125 using

```diff
diff --git a/k8s/helmfile/env/local/wbaas-backup.values.yaml.gotmpl b/k8s/helmfile/env/local/wbaas-backup.values.yaml.gotmpl
index 8d9459c..47cd659 100644
--- a/k8s/helmfile/env/local/wbaas-backup.values.yaml.gotmpl
+++ b/k8s/helmfile/env/local/wbaas-backup.values.yaml.gotmpl
@@ -6,8 +6,15 @@ job:
   successfulJobsHistoryLimit: 1
   cronSchedule: "* * 1 1 *"
 
+scratchDiskSpace: 4Gi
+
+restorePodRunning: false
+
 storage:
-  scratchDiskSpace: 4Gi
-  gcs:
-    bucketName: nacho-cheese
-    uploadToBucket: false
+  bucketName: backups
+  uploadToBucket: true
+  accessKeySecretName: minio-credentials
+  accessKeySecretKey: rootUser
+  secretKeySecretName: minio-credentials
+  secretKeySecretKey: rootPassword
+  endpoint: http://minio.default.svc.cluster.local:9000
diff --git a/k8s/helmfile/helmfile.yaml b/k8s/helmfile/helmfile.yaml
index ed21617..61eb86e 100644
--- a/k8s/helmfile/helmfile.yaml
+++ b/k8s/helmfile/helmfile.yaml
@@ -210,7 +210,7 @@ releases:
 
   - name: wbaas-backup
     namespace: default
-    chart: wbstack/wbaas-backup
+    chart: ./../../../charts/charts/wbaas-backup
     version: 0.0.6
     <<: *default_release
```

k8s/helmfile/env/local/minio.values.yaml.gotmpl
tarrow: I feel like this is fine; not that much space really and even slow laptops may have more ssd

tf/env/local/secrets-minio.tf
deer-wmde: no strong opinion - I think both would be fine

deer-wmde (approved):
worked beautifully locally

PR #1083, opened 2023-08-10
refactor(jobs): add bash -v for verbosity

deer-wmde:
This adds the bash flag `-v` for some verbosity to our k8s yaml job templates, so we can see when and what command actually runs.

From the manpage:
```
       -v        Print shell input lines as they are read.
```

I left the ones with SQL out because it would probably make them print passwords.


deer-wmde:
it really makes no sense as you also don't see the used env var output, closing.

PR #1095, opened 2023-08-14
tf: module-monitoring-19 Group ES metrics by cluster

tarrow:
Grouping by elasticsearch cluster should ensure that there are not erroneous alerts fired about a loss of metric series when the identity of the exporter pod changes


m90 (approved):
Do we have an established workflow for creating such a "tagged release" for a tf module? Asking because I'd have included the changelog update in this PR, but then again I might be missing something.

tarrow:
we don't but probably we should. In any case the lack of changelog here is oversight by me anyway (i'd written it but failed to stage it)

PR #1100, opened 2023-08-14
fix(local): fix adhoc-jobs-podquota name

deer-wmde:
Fixes a typo that breaks local tf conf


m90 (approved):
At least I got one out of three right, ugh.

PR #1103, opened 2023-08-14
tf:both fix platform summary error

tarrow:
It appears that the time time we stored our dashboard the json represention of the dashboard did not return a style object for text objects[0]. Now it always
returns empty ones.

This patch therefore adds them to avoid the perpetual-diff situation.

This patch should be a no-op on application and should remove TF perma-diff

[0] https://cloud.google.com/monitoring/api/ref_v3/rest/v1/projects.dashboards#text


deer-wmde (approved):
can confirm it works YAY

PR #1122, opened 2023-08-24
fix(api): pod ip ranges are incorrect

m90:
I mixed up the values for production and staging.

### DIFF
--- a/k8s/helmfile/env/production/api.values.yaml.gotmpl
+++ b/k8s/helmfile/env/production/api.values.yaml.gotmpl
@@ -120,7 +120,7 @@ app:
     errorReportingEnabled: true
 
 trustedProxy:
-  proxies: ['10.112.0.0/14']
+  proxies: ['10.108.0.0/14']
 
 resources:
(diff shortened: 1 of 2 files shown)


k8s/helmfile/env/production/api.values.yaml.gotmpl
```diff
@@ -120,7 +120,7 @@ app:
     errorReportingEnabled: true
 
 trustedProxy:
-  proxies: ['10.112.0.0/14']
+  proxies: ['10.108.0.0/14']
```
m90, line 123: To prevent this from going out of sync again, should we put them in `base.yaml`?

tarrow: probably "best" to actually source them from TF. `google_container_cluster.wbaas-2.cluster_ipv4_cidr` for example has what we need. We would need to pass this into k8s with a configmap and then consume it in both api and mediawiki

tarrow (approved):

PR #1123, opened 2023-08-24
feat(tf): inject pod ip range through config map

m90:
Companion PR for https://github.com/wbstack/charts/pull/129

### DIFF
--- a/tf/env/local/cluster.tf
+++ b/tf/env/local/cluster.tf
@@ -1,0 +1,9 @@
+resource "kubernetes_config_map" "cluster" {
+  metadata {
+    name = "cluster-config-map"
+  }
+
+  data = {
+    ipv4_range = "10.0.0.0/8"
+  }
+}
--- a/tf/env/local/cluster.tf
+++ b/tf/env/local/cluster.tf
@@ -1,6 +1,6 @@
-resource "kubernetes_config_map" "cluster" {
+resource "kubernetes_config_map" "cluster_config_map" {
   metadata {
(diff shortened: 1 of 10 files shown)


tf/env/local/cluster.tf
```diff
@@ -0,0 +1,9 @@
+resource "kubernetes_config_map" "cluster" {
+  metadata {
+    name = "cluster-config-map"
+  }
+
+  data = {
+    ipv4_range = "10.0.0.0/8"
```
tarrow, line 7: seems easiest to hardcode this here. I guess we could probably pull this from minikube somehow but hardly worth the effort

tarrow (approved):
looks good to me; I wonder if we want to somehow lint tf more aggressively s that we don't see such big diffs as staging/cluster.tf when it's really just whitespace moving

m90:
Yeah we still have this pending: https://github.com/wmde/wbaas-deploy/pull/893 - when merged we could also have a CI action that fails the build when `terraform fmt` does create changes.

tf/env/production/cluster.tf
```diff
@@ -59,3 +59,13 @@ resource "google_container_node_pool" "wbaas-3_highmem-16" {
         max_unavailable = 0
     }
 }
+
+resource "kubernetes_config_map" "cluster" {
```
tarrow, line 63: I believe that only sticking this in the default ns is fine since it should only be needed by web pods not adhoc/api jobs etc.

PR #1140, opened 2023-09-06
chore(istio): add chart version

deer-wmde:
Previously we did not define a chart version here and this got us a surprise upgrade


tarrow:
I guess the `.sh` was accidentally checked in?

PR #1149, opened 2023-09-12
Increase signup limit from 10 to 50 per day on production

rosalieper:
This is a temporal increase for a workshop today.


rosalieper:
merging this because it is already deployed.

PR #1163, opened 2023-09-13
fix: Throttle Signup limit on production

dati18:


m90:
Note that while the change only touches the production file, it will also affect staging, so we need to apply it to both envs after merging.

m90:
I also just double checked and we're now at 3 sign ups in the last 24h, which means this is good to go.

```
>>> User::where('created_at', '>=', \Carbon\Carbon::now()->sub(new \DateInterval('PT24H')))->count()
=> 3
```

PR #1164, opened 2023-09-14
feat: add linting for terraform files

m90:
Enforces ADR No. 9, added in #893

This includes two changes:
- fail the build in case any terraform file is not formatted correctly
- reformat all files that would currently fail that test

When used like this, it'd be left up to the individual dev to decide how to format their changes (i.e. via IDE, CLI or git hooks).

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,5 +1,5 @@
 on: push
-name: Test
+name: Lint
 jobs:
   lint:
     runs-on: ubuntu-latest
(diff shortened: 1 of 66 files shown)


.github/workflows/lint.yaml
```diff
@@ -1,5 +1,5 @@
 on: push
-name: Test
+name: Lint
```
m90, line 2: "Lint" is probably a bit less misleading as we're not really testing anything here.

tarrow (approved):

deer-wmde (approved):

PR #1171, opened 2023-09-19
chore(cleanup): recaptcha values

deer-wmde:
This PR cleans up some value files
  - prod: api: `sitekeySecretName` gets removed (only UI needs it)
  - local & staging: api: recaptcha3 secret secret name gets read from private.yaml 
  - local: `external` block from api.values gets removed - honestly idk why it was even there
  - local: ui: recaptcha3 site secret name gets read from private.yaml
  - local: mw: recaptcha2 secrets name gets read from private.yaml
  - local: rename recaptcha secrets to match staging/prod (`recaptcha-vN-secrets` vs `recaptcha-vN-dev-secrets`)
    - **this will require to update your tfvars files**

### DIFF
--- a/k8s/helmfile/env/local/api.values.yaml.gotmpl
+++ b/k8s/helmfile/env/local/api.values.yaml.gotmpl
@@ -36,20 +36,13 @@ app:
     smtpPasswordSecretKey: password
   recaptcha:
     minScore: 0.5
-    secretSecretName: recaptcha-v3-dev-secrets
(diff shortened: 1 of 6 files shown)


deer-wmde:
> Probably at some point we ought to just give all these secrets the same name in all environments but whatever ;)

yeah I wanted to include that, but it's in the terraform modules... which is the same repo I now realize. Will add

PR #1183, opened 2023-10-04
docs(adr): ADR no. 10 - Backup Bucket Usage

deer-wmde:
A proposal to use GCP buckets differently, to avoid situations where we have zero logical backups.

An example of how this could look like can be currently seen in the staging backup bucket: https://console.cloud.google.com/storage/browser/wikibase-dev-sql-backup

Another approach to tackle this and logical SQL backups in general could be to revisit [ADR no. 1](https://github.com/wmde/wbaas-deploy/blob/9ff399e53b55f6be1939c458ea2003956b7baa72/doc/adr/0001-use-gke-backup-for-maria-logical-backups.md) since meanwhile the feature availability reached GA


deer-wmde:
Concern raised in daily "but is there no risk we quickly overwrite the backup with 7 different nearly empty/corrupt versions?"

deer-wmde:
> Concern raised in daily "but is there no risk we quickly overwrite the backup with 7 different nearly empty/corrupt versions?"

Today I remembered that we have a check for that in place, so maybe this ADR would still be helpful to implement.

https://github.com/wmde/wbaas-backup/blob/main/src/validate_expected_files.sh

tarrow (approved):
let's at least merge this as proposed

PR #1184, opened 2023-10-04
docs(adr): ADR no. 11 Handling FQDNs

deer-wmde:
An ADR about handling FQDNs, especially if they consinst of non-ASCII characters.

### DIFF
--- a/doc/adr/0011-handling-fqdns.md
+++ b/doc/adr/0011-handling-fqdns.md
@@ -1,0 +1,30 @@
+# 11) Handling FQDNs {#adr_0011}
+
+<!--
+Don't forget to update the TOC in index.md when adding a new record
+-->
+
+Date: 2023-10-04
+
+## Status
+
+proposed
+
+## Context
+(FQDN = Fully qualified domain name)
+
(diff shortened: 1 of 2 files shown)


doc/adr/0011-handling-fqdns.md
```diff
@@ -0,0 +1,30 @@
...
+proposed
+
+## Context
+(FQDN = Fully qualified domain name)
+
+Wikibase.cloud allows it's users to create wikis with subdomains on `wikibase.cloud` or to use their own domain names. In both cases, the (resulting) FQDN gets stored in MariaDB database as-is. In August 2023 it became apparent that FQDNs with special characters (= non-ASCII), causes troubles in the system [1], one of which being k8s only allowing handling of hostnames according to RFC 1123 [2][3].
+
+## Decision
+
+To circumvent current and future troubles with non-ASCII domain names, from the moment the system receives the name during creation of a wiki, it gets encoded to punycode[4] (an encoding allowing unicode via ascii representation), and gets handled only in that format internally. As soon as the value leaves the internal API, it gets decoded to it's original representation in unicode.
```
deer-wmde, line 20: What are all the places where the domain names get used? How happy are mediawiki, QS, etc. with non-ascii/punycode FQDNs? ElasticSearch?

deer-wmde: I actually remember now testing only using punycode for the k8s ingress, mediawiki wasn't happy about that, I assume similar results for other services. What I'm a bit worried about are the implications for Wikibase and for example results in the Query Service. Can't tell if this would be problematic or be actually the exact right thing to do.

tarrow (approved):

deer-wmde:
I hate to say it but maybe we should rethink this, see https://github.com/wbstack/api/pull/652#issuecomment-1760085952

tarrow (approved):

PR #1185, opened 2023-10-05
feat(monitoring): alert policy for SQL backup disk usage

deer-wmde:
https://phabricator.wikimedia.org/T348148#9228815

### tf-module-monitoring-20
- Adds an alerting policy in case the scratch disk used by the logical SQL backups is getting to a critical usage (it will fill up eventually again and again, and this way we can act in advance)
- Fixes alerting policy for failed backups by setting the duration period to 24h (previously 24h 30m, which is invalid, max is 24h)


tarrow (approved):
Looks good to me, trust you've already tested on staging so I won't repeat this

deer-wmde:
I pushed another commit after successfully testing that setting the duration period of the other backup alert to 24h (instead of 24h 30m) fixes it

PR #1187, opened 2023-10-10
chore(es6): staging - increase timeout for cluster info update job

deer-wmde:
Temporary staging pendant to #1186


k8s/helmfile/env/staging/elasticsearch.values.yaml.gotmpl
```diff
@@ -10,6 +10,11 @@ imageTag: "6.8.23-wmde.6"
 
 esJavaOpts: "-Xms2g -Xmx2g"
 
+# this section can be removed once #1186 is merged
+esConfig:
+  elasticsearch.yml: |
+    cluster.info.update.timeout: 20s
```
m90, line 16: I think this would need to repeat all that there is in the production config as it is a plain string and won't get merged.

PR #1192, opened 2023-10-13
feat(elasticsearch): use 7.10.2 cluster for local and staging

AndrewKostka:


k8s/helmfile/env/local/api.values.yaml.gotmpl
```diff
@@ -43,6 +43,7 @@ app:
     serviceAccountSecret: null
   stackdriver:
     enabled: false
+  elasticSearchHost: elasticsearch-1-master-hl.default.svc.cluster.local:9200
```
tarrow, line 46: I believe this should probably use the "normal" elasticsearch host rather than specifically the master headless one.

I believe in this context the `hl` version doesn't include the appropriate readiness checks (although we might have turned them off) and also we should probably be aiming at the non-master (coordinating only nodes) anyway (see like the second paragraph of here https://opster.com/guides/elasticsearch/high-availability/coordinating-only-dedicated-coordinating-node/)

tarrow:
You can specify a list: see https://github.com/wbstack/api/pull/624

PR #1196, opened 2023-10-17
chore(elasticsearch): set heapSize to half the memory limit

AndrewKostka:


m90 (approved):
I'm surprised this isn't the default. At least this what we have been doing in v6 and what's recommended basically everywhere else as well.

PR #1227, opened 2023-10-30
chore: update chart for queryservice-updater

m90:
Ticket https://phabricator.wikimedia.org/T333163

Requires https://github.com/wbstack/charts/pull/130

### DIFF
--- a/k8s/helmfile/env/production/api.values.yaml.gotmpl
+++ b/k8s/helmfile/env/production/api.values.yaml.gotmpl
@@ -1,5 +1,5 @@
 image:
-  tag: 8x.25.7
+  tag: 8x.26.0
 
 replicaCount:
   web: 1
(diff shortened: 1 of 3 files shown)


k8s/helmfile/env/production/api.values.yaml.gotmpl
```diff
@@ -44,6 +44,8 @@ ingress:
   {{- end }}
 
 wbstack:
+  qsBatchesMarkFailedAfter: 3
+  qsBatchesRetryAfter: 'PT300S'
```
deer-wmde, line 48: If I'm not mistaken, in the chart these are named starting with `qsBatch` (vs `qsBatches`) so one of them needs to change

deer-wmde: also `RetryAfter` vs `PendingTimeout`

deer-wmde (approved):

PR #1229, opened 2023-10-30
feat(nginx): allow blocking certain user-agents on production

AndrewKostka:


m90:
We do have a blocking mechanism in place already and it's configured here: https://github.com/wmde/wbaas-deploy/blob/main/k8s/helmfile/env/production/ingress-nginx.values.yaml.gotmpl#L11

PR #1231, opened 2023-10-30
fix(tf): sql alerting policy

deer-wmde:
https://phabricator.wikimedia.org/T349817

This PR aggregates the log based metric that is used for the sql backup alerting. 

image

### DIFF
--- a/tf/modules/monitoring/sql-backup-failure-alert.tf
+++ b/tf/modules/monitoring/sql-backup-failure-alert.tf
@@ -28,7 +28,10 @@ resource "google_monitoring_alert_policy" "alert_policy_sql_logical_backup_failu
       aggregations {
         cross_series_reducer = "REDUCE_COUNT"
         per_series_aligner = "ALIGN_COUNT"
-        group_by_fields = ["container_name"]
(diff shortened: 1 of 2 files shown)


tf/modules/monitoring/sql-backup-failure-alert.tf
```diff
@@ -24,6 +24,12 @@ resource "google_monitoring_alert_policy" "alert_policy_sql_logical_backup_failu
     condition_absent {
       filter   = "metric.type=\"logging.googleapis.com/user/${google_logging_metric.sql_logical_backup_success.name}\" AND resource.type=\"k8s_container\""
       duration = "86400s"
+
+      aggregations {
+        cross_series_reducer = "REDUCE_COUNT"
+        per_series_aligner = "ALIGN_COUNT"
+        group_by_fields = "container_name"
```
tarrow, line 31: try something like this:
```suggestion
        group_by_fields = ["container_name"]
```

m90 (approved):
LGTM, do you still wanna add the changelog entry in this PR so we can tag it once merged?

PR #1244, opened 2023-11-08
chore(mw): raise memory requests, lower number of replicas

m90:
Ticket https://phabricator.wikimedia.org/T349887

Raising the requests for the pods lowers the chances of them being evicted by memory pressure on the host node.

Lowering the number of replicas again might help with reducing the number of DB connections being openend simultaneously.


deer-wmde:
adding my approval here retro actively

PR #1259, opened 2023-11-14
chore(elasticsearch): remove 6.8.22 configs and helmfile entry

AndrewKostka:


tarrow (approved):
nice; locally people might forget to apply this intermediate patch so we can remind them in the daily a few times so this zombie release doesn't hang around on their laptops untracked by helmfile

PR #1261, opened 2023-11-14
chore(elasticsearch): reduce CPU requests and limits on production

AndrewKostka:


k8s/helmfile/env/production/elasticsearch-1.values.yaml.gotmpl
```diff
@@ -34,10 +34,10 @@ data:
   heapSize: 32768m
   resources:
     requests:
-      cpu: "10"
+      cpu: "2"
       memory: "64Gi"
     limits:
-      cpu: "10"
+      cpu: "4"
```
m90, line 40: This seems pretty drastic to me right now (even if it does not seem to use more). I wouldn't mind if we maybe start a little more conservative and see if it behaves well still? In the end, we probably also should consider the size of instances we can get as we need to provision a lot of memory. If this means we have spare CPUs doing nothing at all all the time, we might wanna allocate them here.

m90 (approved):

tarrow (approved):

PR #1268, opened 2023-11-15
chore(elasticsearch): startup probe success if cluster is green

AndrewKostka:


tarrow (approved):
looks good once you've minted the new image

PR #1290, opened 2023-11-28
chore(api): allocate more resources to queue pods

m90:
The CreateQueryserviceBatches job is now being OOM killed.


m90:
Self merging this so I can unblock production

PR #1324, opened 2023-12-14
Test SQL alerting policy

deer-wmde:
https://phabricator.wikimedia.org/T349817

**This PR needs to be reverted after testing** 

The purpose of this is to "disable" the sql backup cronjobs by using only a single date in November. We want to see if the change of the alerting policy in #1323 fixes it (this way we can test staging and prod at the same time since we saw strange differences here in the past).

- use `tf-module-monitoring-23` on staging and prod
- "disable" the backup cronjob for staging and prod

### DIFF
--- a/k8s/helmfile/env/staging/wbaas-backup.values.yaml.gotmpl
+++ b/k8s/helmfile/env/staging/wbaas-backup.values.yaml.gotmpl
@@ -1,9 +1,6 @@
 image:
   tag: v0.3.1
 
-job:
-  cronSchedule: "0 0 1 11 0"
-
 scratchDiskSpace: 64Gi
 
 storage:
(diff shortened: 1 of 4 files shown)


k8s/helmfile/env/staging/wbaas-backup.values.yaml.gotmpl
```diff
@@ -1,6 +1,9 @@
 image:
   tag: v0.3.1
 
+job:
```
tarrow, line 4: it fine but this isn't necessary since it would inherit from prod

tarrow (approved):

k8s/helmfile/env/staging/wbaas-backup.values.yaml.gotmpl
deer-wmde: ah true!

PR #1332, opened 2023-12-20
fix(tf-monitoring): k8s labels for api queue have changed

m90:
Platform stats went missing since shipping #1329 as the deployment objects now have different labels, and the query in this module would yield no results.

### DIFF
--- a/tf/modules/monitoring/platform-summary.tf
+++ b/tf/modules/monitoring/platform-summary.tf
@@ -5,7 +5,7 @@ resource "google_logging_metric" "platform-summary-metrics" {
         resource.type="k8s_container"
         resource.labels.cluster_name="${var.cluster_name}"
         resource.labels.namespace_name="default"
-        labels.k8s-pod/app_kubernetes_io/component="queue"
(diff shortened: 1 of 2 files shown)


tf/modules/monitoring/platform-summary.tf
```diff
@@ -5,7 +5,7 @@ resource "google_logging_metric" "platform-summary-metrics" {
         resource.type="k8s_container"
         resource.labels.cluster_name="${var.cluster_name}"
         resource.labels.namespace_name="default"
-        labels.k8s-pod/app_kubernetes_io/component="queue"
+        labels.k8s-pod/app_kubernetes_io/component="queue-default"
```
m90, line 8: Not sure if it would make sense to start using a variable here. Maybe yes, as we'll soon move the job to a dedicated job, changing the name again.

tarrow: I *personally* still think this is fine to hardcode here; sure we'll move it soon but probably repeatedly after that

tarrow (approved):

PR #1347, opened 2024-01-08
chore(elasticsearch): up master node memory to 25% of data node

AndrewKostka:


tarrow (approved):
I *think* this should just fit but it might be a little tight

PR #1352, opened 2024-01-10
feat(api): configure local and staging to store api assets using S3 protocol

m90:
Ticket https://phabricator.wikimedia.org/T354744

### DIFF
--- a/k8s/helmfile/env/local/minio.values.yaml.gotmpl
+++ b/k8s/helmfile/env/local/minio.values.yaml.gotmpl
@@ -13,7 +13,28 @@ persistence:
 buckets:
   - name: backups
     policy: public
+  - name: api-assets
+    policy: public
 
 resources:
   requests:
     memory: 0.5Gi
+
+consoleService:
+  port: "80"
+
+consoleIngress:
+  enabled: true
+  path: /
+  hosts:
+    - minio-console.wbaas.localhost
+  annotations:
+    kubernetes.io/ingress.class: nginx
+
+ingress:
+  enabled: true
+  path: /
+  hosts:
+    - minio.wbaas.localhost
+  annotations:
+    kubernetes.io/ingress.class: nginx
--- a/k8s/helmfile/env/staging/api.values.yaml.gotmpl
+++ b/k8s/helmfile/env/staging/api.values.yaml.gotmpl
@@ -1,5 +1,5 @@
 image:
-  tag: 8x.33.0
+  tag: 8x.34.0
 
 ingress:
   tls:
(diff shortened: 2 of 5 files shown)


m90:
Leaving this a draft until chart and API images are actually released. I tested this by using the chart branch checked out locally and skaffolding `api`

m90:
This is also still missing `values` for staging. I'd probably have to test this while dibsing as I'm not entirely sure which configuration values staging actually wants.

k8s/helmfile/env/local/minio.values.yaml.gotmpl
```diff
@@ -13,7 +13,29 @@ persistence:
...
+    policy: public
 
 resources:
   requests:
     memory: 0.5Gi
+
+consoleService:
+  port: "80"
+
+consoleIngress:
```
m90, line 26: This exposes an admin interface for minio which lets you browse the file tree et. al. It's not strictly necessary but it was helpful for me to get this set up. Not sure if we want to keep it or remove it again. For local, I can see no real harm in leaving it.

tf/env/local/config-map.tf
```diff
@@ -2,4 +2,14 @@ module "wbaas-config" {
...
\ No newline at end of file
+}
+
+resource "kubernetes_config_map" "storage-bucket" {
+  metadata {
+    name = "storage-bucket"
+  }
+
+  data = {
+    gcs_api_static_bucket_name = "api-assets"
```
m90, line 13: We probably could jump through a few hoops and deduplicate this (it's repeated in the mino values file). I don't think it'd be worth it though.

k8s/helmfile/env/staging/api.values.yaml.gotmpl
```diff
@@ -1,5 +1,5 @@
 image:
-  tag: 8x.33.0
+  tag: 8x.34.0
```
deer-wmde, line 3: Is the staging conf missing the `storage` section?

k8s/helmfile/env/local/minio.values.yaml.gotmpl
deer-wmde: I can imagine that this will be helpful in the future as well! works nicely

k8s/helmfile/env/staging/api.values.yaml.gotmpl
m90: Thing is I don't really know the values staging and production want yet, so I was planning to add this while dibsing staging. Sorry for not mentioning it yet.

deer-wmde: ah that sounds reasonable, hard to know how GCP behaves otherwise :+1:

m90:
Tested this successfully in staging.

deer-wmde (approved):

PR #1366, opened 2024-01-18
chore(local+staging): use dedicated gcs hmac key for platform api

m90:
Ticket https://phabricator.wikimedia.org/T354286

Requires https://github.com/wbstack/api/pull/727

### DIFF
--- a/k8s/helmfile/env/staging/api.values.yaml.gotmpl
+++ b/k8s/helmfile/env/staging/api.values.yaml.gotmpl
@@ -1,5 +1,5 @@
 image:
-  tag: 8x.35.0
+  tag: 8x.35.1
 
 ingress:
   tls:
--- a/k8s/helmfile/env/staging/api.values.yaml.gotmpl
+++ b/k8s/helmfile/env/staging/api.values.yaml.gotmpl
@@ -48,3 +48,5 @@ trustedProxy:
 
 storage:
   url: https://storage.googleapis.com/wikibase-dev-static
+  accessKeySecretName: public-assets-hmac-key
+  secretKeySecretName: public-assets-hmac-key
--- a/tf/env/staging/secrets-gcs.tf
+++ b/tf/env/staging/secrets-gcs.tf
@@ -22,24 +22,3 @@ resource "kubernetes_secret" "gcs-hmac-key" {
     "secret-key" = google_storage_hmac_key.dev-backup-upload-key.secret
   }
 }
-
-resource "google_storage_hmac_key" "public-assets-key" {
-  service_account_email = google_service_account.dev-api.email
-}
-
-resource "google_project_iam_member" "public-assets-upload" {
-  role    = "roles/storage.admin"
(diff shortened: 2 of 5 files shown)


tf/env/staging/secrets-gcs.tf
```diff
@@ -22,3 +22,24 @@ resource "kubernetes_secret" "gcs-hmac-key" {
     "secret-key" = google_storage_hmac_key.dev-backup-upload-key.secret
   }
 }
+
+resource "google_storage_hmac_key" "public-assets-key" {
```
m90, line 26: Creating an additional key is not a hard requirement to fix the issue, but it's probably a good idea in terms of helping us to debug ownership problems.

tf/env/staging/secrets-gcs.tf
```diff
@@ -22,3 +22,24 @@ resource "kubernetes_secret" "gcs-hmac-key" {
...
+  service_account_email = google_service_account.dev-api.email
+}
+
+resource "google_project_iam_member" "public-assets-upload" {
+  role    = "roles/storage.admin"
+  member  = "serviceAccount:${google_service_account.dev-api.email}"
+  project = local.project_id
```
tarrow, line 33: I might be totally misunderstanding the google docs here but I think this is probably much broader than we want.

As I understand it this would grant this staging key rights to do things like delete the production backup bucket. I'm looking at how broad it sounds from the [storage docs](https://cloud.google.com/storage/docs/access-control/iam-roles#:~:text=Storage%20Admin%20(roles/storage.admin))

I think this resource:  [google_storage_bucket_iam_member](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket_iam#google_storage_bucket_iam_member) is probably more likely what we want. This is analogous to how we grant rights to the state bucket (see [here](https://github.com/wmde/wbaas-deploy/blob/d84bf06e30368a59eacd4742297cc98bb3cba554/tf/env/production/terraform.tf#L19)) and is probably better than overwriting the whole policy like we do with the backup buckets (see [here](https://github.com/wmde/wbaas-deploy/blob/d84bf06e30368a59eacd4742297cc98bb3cba554/tf/modules/buckets/main.tf#L39)).

I imagine the place this should maybe live is in https://github.com/wmde/wbaas-deploy/blob/d84bf06e30368a59eacd4742297cc98bb3cba554/tf/modules/buckets/main.tf

k8s/helmfile/env/staging/api.values.yaml.gotmpl
```diff
@@ -1,5 +1,5 @@
 image:
-  tag: 8x.34.1
+  tag: 8x.34.2
```
deer-wmde, line 2: I suppose these should be 35.1 now

deer-wmde (approved):

tf/env/staging/secrets-gcs.tf
m90: Fixed via https://github.com/wmde/wbaas-deploy/pull/1372

PR #1371, opened 2024-01-19
feat(jobs): Add wrapper job to start QueryService rebuild

tarrow:


m90:
Not sure if this should part of this script, but there is one thing missing: in certain scenarios, whoever runs this script _might_ want to delete all jobs that exist in the `qs-jobs` namespace as otherwise we'd just apply a spec that is already there and might also already have been finished. I.e.: jobs inserted that way are not guaranteed to run unless we a. clean the namespace or b. change the naming mechanism (right now, the job name is a hash of wiki domain + entity ids for that job)

tarrow:
I feel like this script could perhaps warn if the namespace is not empty and ask for confirmation to continue explaining that this is the case?

m90:
> I feel like this script could perhaps warn if the namespace is not empty and ask for confirmation to continue explaining that this is the case?

Sounds reasonable, yes?

PR #1376, opened 2024-01-22
feat: use s3 based storage driver for api in production

m90:
Ticket https://phabricator.wikimedia.org/T354744

Follows up on https://github.com/wmde/wbaas-deploy/pull/1358

### DIFF
--- a/tf/env/production/buckets.tf
+++ b/tf/env/production/buckets.tf
@@ -1,5 +1,5 @@
 module "production-buckets" {
-  source = "git::ssh://git@github.com/wmde/wbaas-deploy//tf//modules/buckets/?ref=tf-module-buckets-0"
+  source = "git::ssh://git@github.com/wmde/wbaas-deploy//tf//modules/buckets/?ref=tf-module-buckets-3"
   providers = {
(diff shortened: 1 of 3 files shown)


tf/env/production/buckets.tf
```diff
@@ -1,5 +1,5 @@
 module "production-buckets" {
-  source = "git::ssh://git@github.com/wmde/wbaas-deploy//tf//modules/buckets/?ref=tf-module-buckets-0"
+  source = "git::ssh://git@github.com/wmde/wbaas-deploy//tf//modules/buckets/?ref=tf-module-buckets-3"
```
m90, line 2: I'm surprised to see that although versions 0, 1 and 2 of this module existed for a long time now (2 was tagged late 2022), this is still using version 0. Could using 3 here introduce unwanted changes?

m90: Diffing the versions seems to be ok https://github.com/wmde/wbaas-deploy/compare/tf-module-buckets-0...tf-module-buckets-3?w=1 and not introduce any unwanted changes.

deer-wmde (approved):

PR #1378, opened 2024-01-22
fix(tf): resource name is used twice in buckets module

m90:
#1377 contained a mistake where the resource name for the HMAC keys would be duplicated.

### DIFF
--- a/tf/modules/buckets/main.tf
+++ b/tf/modules/buckets/main.tf
@@ -94,7 +94,7 @@ resource "google_storage_bucket_iam_member" "backup-upload" {
   bucket = local.gcs_sql_bucket_backup_name
 }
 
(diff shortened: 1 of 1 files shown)


tf/modules/buckets/main.tf
```diff
@@ -94,7 +94,7 @@ resource "google_storage_bucket_iam_member" "backup-upload" {
   bucket = local.gcs_sql_bucket_backup_name
 }
 
-resource "kubernetes_secret" "gcs-hmac-key" {
+resource "kubernetes_secret" "backup-hmac-key" {
   metadata {
     name = "gcs-hmac-key"
```
tarrow, line 99: we should probably tweak this too otherwise we have two different resources with the same name just one in TF land and one in k8s land

tarrow (approved):

PR #1399, opened 2024-01-30
chore(qs-ui): update image to 1.5 in staging and local

m90:
Ticket https://phabricator.wikimedia.org/T347023

Deploys https://github.com/wbstack/queryservice-ui/pull/237


k8s/helmfile/env/local/queryservice-ui.values.yaml.gotmpl
```diff
@@ -1,3 +1,3 @@
 image:
-  tag: "1.2"
```
m90, line 2: Not sure if there is a specific reason local was two versions behind, but when developing I ran 1.5 just fine, so I would assume it's an oversight.

PR #1405, opened 2024-02-05
chore(monitoring): exclude qs-jobs namespace from kube-state-metrics

m90:
Currently diffs against staging like

```diff
monitoring, kube-prometheus-stack-kube-state-metrics, Deployment (apps) has changed:
...
        containers:
        - name: kube-state-metrics
          args:
          - --port=8080
          - --resources=certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,ingresses,jobs,leases,limitranges,mutatingwebhookconfigurations,namespaces,networkpolicies,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses,validatingwebhookconfigurations,volumeattachments
+         - --namespaces-denylist=qs-jobs
          imagePullPolicy: IfNotPresent
          image: "registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.7.0"
          ports:
          - containerPort: 8080
            name: "http"
...
```


m90:
A different approach would be excluding `jobs` from the collectors https://github.com/prometheus-community/helm-charts/blob/5bd2f828f67aea8861f31433edb9c5c503fe5fcc/charts/kube-state-metrics/values.yaml#L309-L339 although I am not sure if this is to be preferred (as it would also affect other job objects)

tarrow:
Did you verify that locally this change means that we don't get this huge in rush of jobs metrics? I was just looking at doing that but wondered if you did already

m90:
> Did you verify that locally this change means that we don't get this huge in rush of jobs metrics? I

I have to admit I wouldn't know where to look locally. I was planning on doing this in staging.

tarrow:
gotcha, that is probably the far less time wasting way to check it. Want to target this PR just at staging and I'll stick an approve on it an we can "just try it"?

m90:
> Want to target this PR just at staging

Good thinking, done so.

tarrow (approved):

PR #1408, opened 2024-02-06
Integrate ArgoCD Helm charts

dati18:
Bug: [T352990](https://phabricator.wikimedia.org/T352990)

you can now install ArgoCD from Helm but it will also target `staging` and `production`. At the moment we only want it to run on local.


k8s/helmfile/argo-cd.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - ./env/staging/base.yaml
+      - ./env/staging/private.yaml
+  local:
+    kubeContext: minikube-wbaas
+    values:
+      - ./env/local/base.yaml
+      - ./env/local/private.yaml
+
+---
+repositories:
```
m90, line 21: Just out of curiosity: is there a rule for when to declare a repository in the sub-Helmfile and when to do it in the top level one? Right now, we sometimes do the one and sometimes the other.

m90 (approved):
I don't know much about the architecture of Argo, so I cannot tell which parts of the system are being installed, but I can confirm that this installs _something_ in the `argocd` namespace for me locally and diffs cleanly against staging and production.

k8s/helmfile/argo-cd.yaml
tarrow: I think we're wildly inconsistent but...
on a totally clean install because the sub-helmfiles are evaluated first it's necessary to have any repositories mentioned in the sub-helmfile or helm will error when evaluating it.

However you can get away with mentioning it in the top-level one because the repository definitions persist between runs of helm so the *second time* you can put it in either place.

Really it should be in both because sub-helmfiles should be standalone

PR #1414, opened 2024-02-07
feat(monitoring): add alarm when number of available es shards is low

m90:
Ticket: https://phabricator.wikimedia.org/T356254

In staging this created an alarm like this:

image

### DIFF
--- a/tf/modules/monitoring/metric-alarms.tf
+++ b/tf/modules/monitoring/metric-alarms.tf
@@ -11,6 +11,19 @@ locals {
       condition_absent        = "300s"
       min_group_by            = "metric.label.es_cluster"
     },
+    "es-cluster-available-shards-${var.environment}" = {
+      display_name            = "Elasticsearch Cluster available shards"
+      filter                  = "metric.type = \"prometheus.googleapis.com/elasticsearch_node_shards_total/gauge\""
+      comparison              = "COMPARISON_GT"
+      evaluation_missing_data = "EVALUATION_MISSING_DATA_ACTIVE"
+      trigger_count           = 1
+      # Currently there is a hard limit of 800 shards per node set via REST API.
+      # The alarm is expected to trigger on 90% usage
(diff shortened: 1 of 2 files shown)


tf/modules/monitoring/metric-alarms.tf
```diff
@@ -11,6 +11,19 @@ locals {
...
     },
+    "es-cluster-available-shards-${var.environment}" = {
+      display_name            = "Elasticsearch Cluster available shards"
+      filter                  = "metric.type = \"prometheus.googleapis.com/elasticsearch_node_shards_total/gauge\""
+      comparison              = "COMPARISON_GT"
+      evaluation_missing_data = "EVALUATION_MISSING_DATA_ACTIVE"
+      trigger_count           = 1
+      # Currently there is a hard limit of 800 shards per node set via REST API.
+      # The alarm is expected to trigger on 90% usage
+      threshold_value  = 720
```
m90, line 22: Does this need to be configurable so we can use a different value in staging?

m90:
I'm not entirely sure the node count needs to be taken into account here. While the naming is indeed a bit misleading, `elasticsearch_node_shards_total` will give me a single value per node, so I would expect this alarm to trigger when one node reaches the threshold derived from the per-node configuration value. This should work for every possible number of nodes.

Or am I missing something fundamental?

m90:
So I tested this in staging and it does indeed count against the "per node" values. I am not entirely sure if the metric is an average of all values or just the very first one to appear to be honest.

tarrow:
I think it must be that it will alert when any series exceeds the threshold; that at least is the most plausible explanation to me

tf/modules/monitoring/metric-alarms.tf
tarrow: Given this is 800 there too for now I think not

tarrow (approved):

PR #1278, opened 2023-11-21
docs(deploy): switching cluster context

deer-wmde:
This PR adjusts the docs: Switching cluster contexts used to be not needed in order to deploy changes, but nowadays it is.

PR #1429, opened 2024-02-13
fix(k8s): add environment definitions to subhelmfiles

tarrow:
It appears that without these present helmfile uses the default environment (and thus kubecontext) for applying changes due to these subhelm files.

This is a big issue; it means that you may accidentally apply to the wrong environment and there will be no hint


m90 (approved):

deer-wmde:
ah this probably solved this mystery? https://github.com/wmde/wbaas-deploy/pull/1278

PR #1430, opened 2024-02-15
k8s:production upgrade kube-prometheus-stack Chart to 56.6.2

tarrow:
This does not update the GKE prometheus-engine/prometheus image. I propose doing this as a seperate second step after the rest of the stack is present in both staging and production

Bug: T356049

### DIFF
--- a/k8s/helmfile/helmfile.yaml
+++ b/k8s/helmfile/helmfile.yaml
@@ -207,7 +207,7 @@ releases:
 
   - name: kube-prometheus-stack
     namespace: monitoring
-    version: '{{ if eq .Environment.Name "production" }} 45.1.1 {{ else }} 56.6.2 {{ end }}'
+    version: '56.6.2'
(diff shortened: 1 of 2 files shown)


m90:
Did we already figure out the mystery diff in staging coming from this update?

Right now I still get 

```diff
monitoring, kube-prometheus-stack-operator, ClusterRole (rbac.authorization.k8s.io) has changed:
...
    - storage.k8s.io
    resources:
    - storageclasses
    verbs:
    - get
- - apiGroups:
-   - discovery.k8s.io
-   resources:
-   - endpointslices
-   verbs:
-   - get
-   - list
-   - watch
```

m90 (approved):

tarrow:
We did not figure out this diff. I was hoping that once we also upgrade the GKE prometheus-engine/prometheus image this will go away. If it doesn't they we have to do some investigating

PR #1441, opened 2024-02-23
chore(elasticsearch): use a dedicated node pool for data nodes

AndrewKostka:


tf/env/production/cluster.tf
```diff
@@ -59,3 +59,50 @@ resource "google_container_node_pool" "wbaas-3_highmem-16" {
...
+    service_account = "default"
+    shielded_instance_config {
+      enable_integrity_monitoring = true
+      enable_secure_boot          = false
+    }
+    logging_variant = "MAX_THROUGHPUT"
+    taint = [
+      {
+        key    = "app.kubernetes.io/component"
+        value  = "data"
```
tarrow, line 95: Yeah, we think this will work!

tarrow (approved):
LGTM, let's try to hear something from someone who can approve spending before we actually do this though see the main wikibase.cloud mattermost

PR #1442, opened 2024-02-23
chore(elasticsearch): reduce memory for data nodes

AndrewKostka:


tarrow (approved):
looks fine; did you run the number and check 4GB is enough space for these to fit on a 64GB node after all the other overheads?

tarrow (approved):

PR #1461, opened 2024-02-29
chore(elasticsearch): use a dedicated node pool for master nodes

AndrewKostka:


tf/env/production/cluster.tf
```diff
@@ -60,6 +60,57 @@ resource "google_container_node_pool" "wbaas-3_highmem-16" {
...
+      "wbaas/pool" = "search-master"
+    }
+    taint = [
+      {
+        key    = "wbaas/pool"
+        value  = "search-master"
+        effect = "NO_SCHEDULE"
+      }
+    ]
+    logging_variant = "MAX_THROUGHPUT"
```
tarrow, line 101: we might need to toggle this if really can't make stuff fit with fluentbit running here

tarrow (approved):

PR #1470, opened 2024-03-05
Revert "chore(cluster): delay cluster update for production"

AndrewKostka:
Reverts wmde/wbaas-deploy#1459


tarrow (approved):
and we're now doing what we said we might do by constantly moving it

PR #1471, opened 2024-03-05
chore(cluster): prepare a smaller node pool for production

AndrewKostka:


tarrow (commented):
I was sort of expecting us to be shrinking `n2-highmem-16-pool` before adding yet more nodes, is that not the case?

Once the search-master pool is used does that not give us enough room to shrink down the number of old generic nodes?

tarrow (approved):

PR #1475, opened 2024-03-06
chore(cluster): reduce node count to 0 on production

AndrewKostka:


tarrow:
could maybe delete the whole resource at this point

PR #1503, opened 2024-03-18
chore(api): update chart in local and staging to include service monitor object

m90:
Ticket https://phabricator.wikimedia.org/T342866

### DIFF
--- a/k8s/helmfile/helmfile.yaml
+++ b/k8s/helmfile/helmfile.yaml
@@ -152,7 +152,7 @@ releases:
 
   - name: api
     chart: wbstack/api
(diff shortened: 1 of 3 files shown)


k8s/helmfile/helmfile.yaml
```diff
@@ -152,7 +152,7 @@ releases:
 
   - name: api
     chart: wbstack/api
-    version: "0.31.0"
+    version: "0.31.1"
```
m90, line 155: This is a noop as long as the serviceMonitor is not explicitly enabled, so it's fine to also ship this to production for now.

PR #1505, opened 2024-03-19
chore: Switch to OpenTofu

tarrow:
This PR attempts to migrate references to terraform to opentofu. References to terraform remain:
- in links to docs that don't currently have an obvious analogue
- in ADRs which should be immutable
- some variable names that may be better to migrate more carefully

I agreed with the comment on the ticket that it seems to just-work fine now.

Bug: T348311


m90:
Are there any breaking changes in this that hard require using `tofu`?

I'm wondering if we should soft start by falling back to a `terraform` executable if `tofu` is not found, or if this a bad incentive for not installing `tofu` in the first place?

tarrow:
IMHO this is a bad incentive for not properly moving and in the event we start to find there is a tiny drift will just add confusion. At the very least I see this as potentially causing fighting over the changes in `tf/env/production/.terraform.lock.hcl` and friends

m90:
Ok, then let's not do it. I agree with your concerns. Fine with merging this as is whenever we feel ready.

m90 (approved):

PR #1513, opened 2024-03-26
Deploy ArgoCD to staging

deer-wmde:
https://phabricator.wikimedia.org/T360874

This PR adjusts the environment condition, which now evals to true for `local` and `staging` but not `production`.

It also introduces values files for the different envs with the default resource values suggested as comments in the root values file: https://raw.githubusercontent.com/argoproj/argo-helm/38900afada6cbc51e469d53fbf20ebba6f1cb07b/charts/argo-cd/values.yaml


m90:
This is a bit twisted, but if I understand the ticket (as well as @tarrow) correctly, we want to deploy Argo to the production cluster only, while skipping staging. The idea behind it is that having a single argo deployment for both environments will allow us to use autopromotion at some point.

deer-wmde:
> This is a bit twisted, but if I understand the ticket (as well as @tarrow) correctly, we want to deploy Argo to the production cluster only, while skipping staging. The idea behind it is that having a single argo deployment for both environments will allow us to use autopromotion at some point.

I see, thanks for pointing that out. Seems like I missed a lot of discussion around this topic

m90:
> Seems like I missed a lot of discussion around this topic

I only just understood this myself yesterday.

deer-wmde:
Shouldn't we still first deploy this thing to staging to figure out how to configure & use it?

k8s/helmfile/env/production/argo-cd-base.values.yaml.gotmpl
```diff
@@ -0,0 +1,76 @@
...
+redis:
+  resources:
+   limits:
+     cpu: 200m
+     memory: 128Mi
+   requests:
+     cpu: 100m
+     memory: 64Mi
+
+  exporter:
```
tarrow, line 51: is this exporter sidecar even enabled? by default I thought it was turned off?

Also if we turn it on we probably (in a follow up) want to add some Prometheus object to actually scrape it

deer-wmde: Ah you're right, it's disabled anyway per default

tarrow (approved):

PR #1519, opened 2024-04-02
chore(qs-gateway): update image to 2.3 and update chart

m90:
Ships https://github.com/wbstack/queryservice-gateway/pull/164 and https://github.com/wbstack/charts/pull/142

### DIFF
--- a/k8s/helmfile/helmfile.yaml
+++ b/k8s/helmfile/helmfile.yaml
@@ -183,7 +183,7 @@ releases:
   - name: queryservice-gateway
     namespace: default
     chart: wbstack/queryservice-gateway
-    version: 0.1.3
+    version: 0.2.0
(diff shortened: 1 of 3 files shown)


k8s/helmfile/helmfile.yaml
```diff
@@ -183,7 +183,7 @@ releases:
   - name: queryservice-gateway
     namespace: default
     chart: wbstack/queryservice-gateway
-    version: 0.1.3
+    version: 0.2.0
```
m90, line 186: I would think it's ok to set this for production as well without updating the image just yet as all it does is set an extra env var that is not picked up by `2.2`

tarrow: yeah fine; I think you can self merge this if you wanted

PR #1521, opened 2024-04-04
chore(elasticsearch): recover from a cold start

AndrewKostka:


tarrow (approved):

tarrow:
We need to try down *why* this cold start happened too; all the pods seem to have been deleted at the same time but not the STS

PR #1533, opened 2024-04-12
feat: add Tailscale for Elasticsearch on staging

AndrewKostka:


k8s/helmfile/tailscale-operator.yaml
```diff
@@ -0,0 +1,10 @@
+repositories:
```
m90, line 1: Sub-Helmfiles will unfortunately need to repeat this block: https://github.com/wmde/wbaas-deploy/blob/097a6bf415eccd37843397ab336e910d373e7f8c/k8s/helmfile/istio-control-plane.yaml#L1-L18, else it's possible to accidentally target different environments when running `make apply-env` in certain scenarios.

tf/env/staging/namespaces.tf
```diff
@@ -66,3 +66,9 @@ resource "kubernetes_resource_quota" "qs-jobs-podquota" {
     scopes = ["BestEffort"]
   }
 }
+
+resource "kubernetes_namespace" "tailscale-namespace" {
```
m90, line 70: IIRC helmfile will do this for us when deployng the tailscale operator the tailscale namespace. The namespaces we create in terraform are currently for use cases where there's no release tied to the namespace, e.g. for running jobs.

k8s/helmfile/tailscale-operator.yaml
```diff
@@ -0,0 +1,10 @@
...
+  - name: tailscale
+    url: https://pkgs.tailscale.com/helmcharts
+
+releases:
+  - name: tailscale
+    namespace: tailscale
+    chart: tailscale/tailscale-operator
```
m90, line 8: In order to be able to limit resource usage (i.e. limits/request on CPU and RAM), we also still need a values file for this chart.

tf/env/staging/namespaces.tf
m90: Ah, got it.

m90 (approved):

PR #1543, opened 2024-04-18
feat: add Tailscale for Elasticsearch on production

AndrewKostka:


m90 (approved):
We tested this on staging with my personal setup, where it worked very well, I think this is good to go.

PR #1545, opened 2024-04-19
argocd core configuration

deer-wmde:
These are the first bits of configuration for argocd. There are no app deployments involved in this PR.


Makefile
```diff
@@ -11,7 +11,7 @@ help:
 .PHONY: minikube-start
 minikube-start: # @HELP Start a local k8s cluster using minikube
 minikube-start:
-	minikube --profile minikube-wbaas start --kubernetes-version=1.26.5
+	minikube --profile minikube-wbaas start --kubernetes-version=1.26.14
```
tarrow, line 14: comment: Makes sense to bump but could perhaps could come in a separate PR. In any case we actually have split 1.26 and 1.27 in our GKE clusters right now

Makefile
```diff
@@ -43,6 +43,11 @@ minikube-tunnel: # @HELP Open a tunnel to the local cluster and expose it on an
 minikube-tunnel:
 	./bin/minikube-tunnel
 
+.PHONY: argocd-port-forward
+argocd-port-forward: # @HELP Forward port 8080 of argocd-server to access it on the host system
+argocd-port-forward:
```
tarrow, line 48: suggestion: scope this to a specific environment. I'm not sure if this is intended to be for `local` `staging` or `production` or all of them. At the moment we are mostly keeping the `make` targets constant independent of what the user currently has their kubecontext set to

tarrow: *If* this is just for `local` why not add a special ingress like we do for mailhog?

bin/get-argocd-password
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+kubectl -n argocd get secrets argocd-initial-admin-secret -o jsonpath='{.data.password}' | base64 --decode
```
tarrow, line 2: complement: nice, I bet I'll forget this is here but super convenient

k8s/helmfile/argo-cd.yaml
```diff
@@ -1,37 +1,18 @@
-environments:
-  default:
```
tarrow, line 2: issue: I think this block actually needs to remain otherwise even if you (for example) `helmfile apply -e local` you will actually target whatever is the current kube-context

k8s/helmfile/env/local/argo-cd-base.values.yaml.gotmpl
```diff
@@ -1 +1,18 @@
-# local uses production configuration so this is an empty file
+configs:
+  params:
+    server.insecure: "true"
```
tarrow, line 3: question: why set this true but not for the other environments? In any case trying to access it after port forwarding to localhost:8080 will probably result in a TLS error right?

doc/deployments/argocd.md
```diff
@@ -0,0 +1,18 @@
...
+
+1. Enable admin access by flipping `admin.enabled` to `true` in our values file (or manually in the ConfigMap `argocd-cm`)
+    - this is enabled for local clusters by default
+2. Forward port `8080` of the pod `argocd-server`
+    - use Makefile target `argocd-port-forward`
+    - or `kubectl -n argocd port-forward deployments/argo-cd-base-argocd-server 8080`
+3. Visit https://localhost:8080/ and log in.
+    - for local clusters TLS is disabled: http://localhost:8080/
+
+### Security considerations
```
tarrow, line 17: suggestion: phrase this to make it clear to future us who needs to get admin access doesn't need to act on this at the time.

suggestion: I believe it would also be an option to have us set this password fresh each time we need it?

tarrow (changes requested):
Looks like a good start; thanks for this! Couple of thoughts :)

k8s/helmfile/env/local/argo-cd-base.values.yaml.gotmpl
deer-wmde: just for convenience. The cert changes rapidly for local and hopefully never on the other envs, so you can add exceptions for staging and prod

doc/deployments/argocd.md
deer-wmde: I think that would require us to use the argocd cli tool on our machines. At least I haven't had luck to do this from within the argo pods yet

Makefile
deer-wmde: > If this is just for local why not add a special ingress like we do for mailhog?
I thought of that as well, but I haven't looked into it yet. Might be best to remove this port forwarding target alltogether

doc/deployments/argocd.md
deer-wmde: I adjusted the bash script to not read the password from the secret, but to generate a new one and write it to the secret

k8s/helmfile/env/local/argo-cd-base.values.yaml.gotmpl
tarrow: ah cool! i hadn't thought of how quickly they refresh!

doc/deployments/argocd.md
```diff
@@ -0,0 +1,12 @@
+# Argo CD
+## Admin access
+> Caution! Admin access should only happen in rare circumstances (testing/diagnosing, for example)
+> as we want to maintain the configuration for everything in git.
+
+1. Reset the admin password by running `./bin/get-argocd-password`
+1. Enable admin access by flipping `admin.enabled` to `true` in our values file (or manually in the ConfigMap `argocd-cm`)
```
tarrow, line 7: this step nicely triggers the restart of the pods to load the new password

tarrow (approved):
Woo LGTM let's gooooooo!

k8s/helmfile/env/local/argo-cd-base.values.yaml.gotmpl
deer-wmde: I should be more clear: I think it only happens once per initial argo setup, but that can be quite often the case locally (especially currently)

PR #1555, opened 2024-04-26
feat(elasticsearch): create elasticsearch-2 cluster on production

AndrewKostka:


tarrow (approved):
Sounds good; there's no reason to have more than 2 right now given we have just one index and two shards (primary/replica)

PR #1556, opened 2024-04-26
feat(elasticsearch): switch to shared Elasticsearch on production

AndrewKostka:


tarrow (approved):
looks good; I assume the empty string removal in-staging works out to a no-op

PR #1569, opened 2024-05-03
local: add statsd-prometheus exporter for MW

tarrow:
Bug: T364107


tarrow:
No point in merging this until we have a new enough version of mediawiki to use it (probably 1.41+)

PR #1573, opened 2024-05-06
chore(elasticsearch): reduce data node resources

AndrewKostka:


tarrow (approved):
cool; happy to see us keep trying to shrink this to what we need; are you keeping some eye on if shrinking this is resulting in slower queries or something?

PR #1584, opened 2024-05-07
chore(elasticsearch): update data node resources

AndrewKostka:


tarrow (approved):
I assume this was to match what we discussed in the workshop?

PR #1585, opened 2024-05-07
docs(adr): 12 stop versioning terraform modules with git

tarrow:


m90:
One idea that's floating around in my head is having symlinks for what we reference in the tf files themselves so we don't have to update things there again and again. Then again, this maybe makes it more visible what's going on. Not sure right now.

In any case, this is definitely an improvement of the current workflow.

tarrow:
That certainly sounds an interesting idea, it might make the diffs somewhat clearer.

I would suggest we try the really simple "duplicate, edit, delete" pattern and then think about symlinking if we find the simple pattern to be insufficient

PR #1596, opened 2024-05-13
chore(elasticsearch): use elasticsearch-2-data-hl service

AndrewKostka:


tarrow (commented):
I'm not sure if we'd want to do this rather than altering the non-headless service.

I wasn't actually certain if Zbyszko was suggesting we needed to change this. I thought he'd said after asking all these questions that it was probably OK because any of the nodes can respond to queries and of the three nodes backed by the service only one was the actual master and the other two were basically idle master eligible nodes.

I believe this service includes non-ready pods which we wouldn't want to be shipping traffic to in any case.

```
 publishNotReadyAddresses: true
```

is set on the `hl` variants

PR #1599, opened 2024-05-14
tf: fix breaker gauge for Elasticsearch

AndrewKostka:


tarrow (approved):
Looks good to me; it's confusing that the auto generated dashboard doesn't work as expected.

Frankly it's a bit odd that `elasticsearch_breakers_tripped` is a counter (https://github.com/prometheus-community/elasticsearch_exporter) but the auto generated dashboard seems to be treating it like a gauge.

Maybe this has different behaviour post 7.10.2 ?

PR #1600, opened 2024-05-14
tf: update Elasticsearch alerting

AndrewKostka:


tf/modules/monitoring/metric-alarms.tf
```diff
@@ -1,28 +1,26 @@
...
       evaluation_missing_data = "EVALUATION_MISSING_DATA_ACTIVE"
       trigger_count           = 1
-      threshold_value         = 1
-      duration                = "60s"
+      threshold_value         = 0
+      duration                = "900s"
       condition_absent        = "300s"
       min_group_by            = "metric.label.es_cluster"
     },
-    "es-cluster-available-shards-${var.environment}" = {
```
tarrow, line 14: I'm happy to see this go; I wonder if we ought to add (in a follow up so not blocking) an alert for the shard size growing beyond a certain size. That would be a comparable alert to this one we are removing for growth in search content we may not have noticed

tf/modules/monitoring/metric-alarms.tf
```diff
@@ -1,28 +1,26 @@
...
+      filter                  = "metric.type = \"prometheus.googleapis.com/elasticsearch_cluster_health_status/gauge\" AND metric.labels.color = \"yellow\""
+      comparison              = "COMPARISON_GT"
       evaluation_missing_data = "EVALUATION_MISSING_DATA_ACTIVE"
       trigger_count           = 1
-      threshold_value         = 1
-      duration                = "60s"
+      threshold_value         = 0
+      duration                = "900s"
       condition_absent        = "300s"
       min_group_by            = "metric.label.es_cluster"
```
tarrow, line 12: I was somewhat struggling to confirm my suspicion but I suspect after inverting the `COMPARISON_LT` to `COMPARISON_GT` this should become `max_group_by` (and same logic for the `red` check)

tarrow: Look's like the provider has undergone a major version bump beyond where we are at (5.29.1) and then looking here https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_alert_policy is seems like the format may have changed somewhat in general for grouping to become more standard with the other apis and we might now need to us the https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_alert_policy#cross_series_reducer for this logic.

tarrow (commented):
Generally looks good; I think we need to take a second look at grouping and it might be easier to do this after updating the google TF provider

tarrow (approved):

tarrow:
> This patch is a no-op on production and will make these changes on staging

whoops, this was wrong - I'd not clocked the final commit in this PR also touched production

PR #1605, opened 2024-05-16
chore: update Tailscale to 1.66.3

AndrewKostka:


tarrow (approved):
LGTM; probably no need to split this into staging and prod variants.

Happy to see they finally released this updated version of the operator too

PR #1608, opened 2024-05-21
docs: document queryservice pipeline and debugging approaches

m90:
I could go into much further detail here, but I feel that might stop being helpful and start being confusing.

Maybe an unbiased reviewer can rate this doc on a helpfulness scale so I can extend / shrink it accordingly.

### DIFF
--- a/doc/debugging-queryservice-backpressure.md
+++ b/doc/debugging-queryservice-backpressure.md
@@ -1,0 +1,119 @@
+# Debugging Backpressure in the QueryService Updater
+
+This document is intended to give a general overview of the existing pipeline which feeds MediaWiki updates into the QueryService, as well as guidance on what to do when the queue is blocked or malfunctioning.
+
+## Lifecycle of an update
+
+When an update in MediaWiki is happening, the following steps will be taken until it's present in the QueryService:
+
+1. MediaWiki calls the Platform API's `/event/pageUpdate` endpoint, pushing information about the change. The Platform API persists these changes in the `event_page_updates` table.
+2. The Platform API will repeatedly run the `CreateQueryserviceBatchesJob` which will:
+	1. Filter all events from the `event_page_updates` table and select those with relevant namespace IDs (i.e. entities, properties, lexemes)
+	2. Group these events by wiki domain and create sets of batches for each wiki. If a batch for a wiki already exists, try to append new items to these batches if possible.
+3. The QueryService Updater repeatedly polls the `/qs/getBatches` endpoint exposed by the Platform API. When a batch is dispensed by the Platform API, it is marked as pending and won't be returned unless this marker is removed again. The QueryService Updater will now feed the batch's items into the QueryService itself.
+4. When the QueryService Updater is done processing a batch, it will signal success or failure to the Platform API which will then update the Batch accordingly.
+5. Batches that failed processing will be retried at most 3 times before being marked as failed permanently.
+6. Already processed batches and page updates that are older than a month will be pruned from the database in cleanup jobs.
+
+## Possible failure scenarios
+
+Currently, feature health is measured by counting the number of unprocessed batches in the Platform API.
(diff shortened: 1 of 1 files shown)


doc/debugging-queryservice-backpressure.md
```diff
@@ -0,0 +1,119 @@
+# Debugging Backpressure in the QueryService Updater
+
+This document is intended to give a general overview of the existing pipeline which feeds MediaWiki updates into the QueryService, as well as guidance on what to do when the queue is blocked or malfunctioning.
+
+## Lifecycle of an update
+
+When an update in MediaWiki is happening, the following steps will be taken until it's present in the QueryService:
+
+1. MediaWiki calls the Platform API's `/event/pageUpdate` endpoint, pushing information about the change. The Platform API persists these changes in the `event_page_updates` table.
```
tarrow, line 9: I think we could add some hint that this is special wbstack secret sauce and not some built int mediawiki thing

doc/debugging-queryservice-backpressure.md
```diff
@@ -0,0 +1,119 @@
...
+	2. Group these events by wiki domain and create sets of batches for each wiki. If a batch for a wiki already exists, try to append new items to these batches if possible.
+3. The QueryService Updater repeatedly polls the `/qs/getBatches` endpoint exposed by the Platform API. When a batch is dispensed by the Platform API, it is marked as pending and won't be returned unless this marker is removed again. The QueryService Updater will now feed the batch's items into the QueryService itself.
+4. When the QueryService Updater is done processing a batch, it will signal success or failure to the Platform API which will then update the Batch accordingly.
+5. Batches that failed processing will be retried at most 3 times before being marked as failed permanently.
+6. Already processed batches and page updates that are older than a month will be pruned from the database in cleanup jobs.
+
+## Possible failure scenarios
+
+Currently, feature health is measured by counting the number of unprocessed batches in the Platform API.
+A healthy system will rarely see this value go above a lower two-digit number.
```
tarrow, line 21:
```suggestion
A healthy system will rarely see this value go above 20.
```

I know it's tricky to put a hard number here but to me this is far easier for my brain to parse

doc/debugging-queryservice-backpressure.md
```diff
@@ -0,0 +1,119 @@
...
+If it goes above this, it could be caused by the following:
+
+### Failures in Platform API
+
+If the Platform API is ingesting events, but fails to dispense them to the updater, the number of batches grows continuously
```
tarrow, line 26:
```suggestion
If the Platform API is ingesting events, but fails to have them completed by the updater, the number of batches grows continuously
```

tarrow, line 26:
```suggestion
If the Platform API is ingesting events, but fails to communicate properly with the updater, the number of batches grows continuously
```

tarrow: I struggled with this phrasing but point 1 and 3 below seem to be cases where the batches are being "dispensed" to the updater but something else is going wrong in the communication

m90: This is not really what I'm trying to communicate here (pun intended): we've seen cases where the Platform API would just return `[]` even when there were batches available (caused by malformed data).

tarrow: Hrm.. certainly point 3 from below isn't a failure to dispense though; maybe we should add this `[]` situation as case 4 below?

doc/debugging-queryservice-backpressure.md
m90: How about `a value in the range between 15 and 50`?

m90:
@tarrow I addressed the points you raised, feel free to have another look.

tarrow (approved):
Awesome; thanks for writing up these docs; they're very readable to me

PR #1619, opened 2024-05-28
chore(tf): add compute-pool-1 on staging

AndrewKostka:


tarrow:
Just to check I understand the chain this can be summed up as:
- move to a similar node pool layout to production
- add large generic compute pool
- remove special nodes for ES master and data
- remove old small generic compute pool
right?

PR #1624, opened 2024-05-29
chore(elasticsearch): lift node restrictions on staging

AndrewKostka:


tarrow (approved):
Approved then you can match the dibs

PR #1626, opened 2024-05-31
chore(tf): add compute-pool-2 on production

AndrewKostka:


tf/env/production/cluster.tf
```diff
@@ -58,3 +58,39 @@ resource "google_container_node_pool" "wbaas-3_compute-pool-1" {
...
+      "https://www.googleapis.com/auth/servicecontrol",
+      "https://www.googleapis.com/auth/trace.append",
+    ]
+    preemptible     = false
+    service_account = "default"
+    shielded_instance_config {
+      enable_integrity_monitoring = true
+      enable_secure_boot          = false
+    }
+    logging_variant = "MAX_THROUGHPUT"
```
m90, line 90: Nothing to do here, but at some point in the future, we should find out whether we really need this. (I think we don't and it just gobbles up resources).

PR #1632, opened 2024-06-04
feat(k8s): add redirect for status page

m90:
## Link to Phabricator

Ticket https://phabricator.wikimedia.org/T365494

### Prior discussion

Some discussion around what status code to use happened on Mattermost https://mattermost.wikimedia.de/swe/pl/hdmn83gjrbbquf4911pb3k5hkc

## Describe the changes

This PR adds a chart that lets users define a list of from/to pairs that will create top level ingresses that redirect to the given URL.

Currently, the only use case is to redirect (temporarily) `status.wikibase.cloud` to `https://wmde.github.io/wikibase-cloud-status/`.

A few questions:

- Does this really need to be so complex? Is there a simpler way to do this, e.g. by configuring the `ingress-nginx/ingress-nginx` chart correctly?
- In the long run, does it even make sense to deploy this to staging?

### DIFF
--- a/charts/redirects/.helmignore
+++ b/charts/redirects/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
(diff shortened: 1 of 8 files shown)


m90:
For testing purposes, I left the ingress dangling in staging:

```
➜  wbaas-deploy git:(main) curl -I https://status.wikibase.dev                                                                                                                                  wbaas-local
HTTP/2 302 
date: Tue, 04 Jun 2024 13:05:20 GMT
content-type: text/html
content-length: 138
location: https://wmde.github.io/wikibase-cloud-status/
```

deer-wmde (approved):

deer-wmde:
> Does this really need to be so complex? Is there a simpler way to do this, e.g. by configuring the ingress-nginx/ingress-nginx chart correctly?
There might be, but I couldn't think of or find one quickly. I think this is great as it is already.

> In the long run, does it even make sense to deploy this to staging?
I guess it's harmless, plus we can use it for testing purposes in case the redirect will change in the future

PR #1639, opened 2024-06-13
docs(elasticsearch): document alias flow

AndrewKostka:


doc/search.md
```diff
@@ -0,0 +1,95 @@
...
+  indexConfig --> Elasticsearch
+```
+
+Before creating the shared index, you will need to have created at least one Wikibase instance. Then it is as simple as running the `elasticSearchInitJob.sh` script found in `k8s/helmfile/jobs`. This only needs to be done once, after creating a new Elasticsearch cluster.
+
+Configuration:
+
+- `WBS_DOMAIN`: domain of any existing Wikibase
+- `CLUSTER_NAME`: Elasticsearch cluster's name (as configured for MediaWiki)
+- `MW_CIRRUSSEARCH_INDEX_BASE_NAME`: shared index's prefix (as configured for Platform API)
```
deer-wmde, line 34: just for my understanding, this value is just something to look up in the DB after creating the first wiki?

deer-wmde:
thanks for the docs, looks very useful!

deer-wmde (approved):

PR #1657, opened 2024-06-25
fix(redirects): later chart versions disable server snippets by default

m90:
The `status.wikibase.cloud` redirect was broken since merging #1638. It seems this was caused by including https://github.com/kubernetes/ingress-nginx/pull/10393, silently removing the required annotation from the ingress.

This PR enables the setting again, restoring behavior as of #1632

### DIFF
--- a/k8s/helmfile/env/production/ingress-nginx.values.yaml.gotmpl
+++ b/k8s/helmfile/env/production/ingress-nginx.values.yaml.gotmpl
@@ -1,4 +1,5 @@
 controller:
+  allowSnippetAnnotations: true
   service:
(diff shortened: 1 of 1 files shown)


tarrow (approved):
looks good to me; to me this is a great example of why relying on other people's definition of breaking change is always a little risky :)

m90:
The interesting part is that the change is marked as breaking, people just couldn't be bothered to bump the version number accordingly.

tarrow:
yeah, very interesting. I'm sure I probably inadvertently do that same thing sometimes

PR #1660, opened 2024-06-26
feat(superset): deploy a lightweight version on staging

AndrewKostka:


m90:
One question from my end: Superset does have user management, so it would also be possible to expose it to the public internet (by adding an ingress like `superset.wikibase.dev`) and then create a dedicated superset user for everyone that needs access. This would mean we can grant access to semi-technical folk without having to use Tailscale and such. Also, it gives us some sort of insights of who's changed what in Superset itself.

Are there any strong reasons for not using the built in access control and instead use our own?

tarrow:
I'm somewhat reading between the lines here but my assumption was that the direction this patch was going in is a kind of defence in depth style: we would also generate individual users for superset (and this just contains some admin bootstrapping user) but we're going to have an initial layer of protection from tailscale since superset would be a very valuable service to compromise (it has access to user account details etc.)

Is that what you were thinking @AndrewKostka ?

tarrow:
Cool, are you expecting this chart to *only* go as far as staging then? If that's the case I'll review it with that in mind

charts/superset/values.yaml
```diff
@@ -0,0 +1,57 @@
...
+  pullPolicy: IfNotPresent
+  tag: 4.0.1
+
+storageSize: 5Gi
+
+secretName: superset-secrets
+secretKeySecretKey: secret
+adminPasswordSecretKey: admin-password
+
+entrypoint: |
```
tarrow, line 12: What was the thinking behind sticking this in the values file rather than in one of the templates?

charts/superset/templates/sqlite-pvc.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ include "superset.fullname" . }}-sqlite-pvc
+  labels:
+    {{- include "superset.labels" . | nindent 4 }}
+spec:
+  persistentVolumeReclaimPolicy: Retain
```
tarrow, line 8: I can't apply the chart due to this:
```
  Error: Failed to render chart: exit status 1: Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(PersistentVolumeClaim.spec): unknown field "persistentVolumeReclaimPolicy" in io.k8s.api.core.v1.PersistentVolumeClaimSpec, ValidationError(PersistentVolumeClaim.spec): unknown field "storageClass" in io.k8s.api.core.v1.PersistentVolumeClaimSpec]
  Error: plugin "diff" exited with error
```

I wonder if you managed to? I think this key is only valid on the PV not the PVC but in any case you might rather add this to the storageclass (or instead use a stateful set in place of the deployment). See this S/O post for some thoughts https://stackoverflow.com/questions/50458165/kubernetes-how-to-do-dynamic-persistentvolumeclaim-with-persistentvolumereclaim

charts/superset/templates/deployment.yaml
```diff
@@ -0,0 +1,93 @@
+apiVersion: apps/v1
+kind: Deployment
```
tarrow, line 2: As per the other comment I wonder if this would better be a statefulset?

tarrow (changes requested):
Couple of minor thoughts but also a blocker in that I wasn't able to apply this chart locally. I'd also like to talk a look at the thing after I can apply it locally and see it all wired up

charts/superset/templates/statefulset.yaml
```diff
@@ -0,0 +1,104 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ include "superset.fullname" . }}
+  labels:
+    {{- include "superset.labels" . | nindent 4 }}
+spec:
```
tarrow, line 7: This won't apply without a `serviceName`.

```
STDERR:
  Error: Failed to render chart: exit status 1: Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1.StatefulSetSpec
  Error: plugin "diff" exited with error
```

charts/superset/values.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+    superset db upgrade
+    superset init
+  fi
+
+  superset fab reset-password \
+    --username cloud \
+    --password $ADMIN_PASSWORD
+
+  pip install mysqlclient==2.1.0
```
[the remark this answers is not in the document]
tarrow: fair enough!

charts/superset/templates/statefulset.yaml
tarrow: Patch from me to make this apply:
```
diff --git a/charts/superset/templates/statefulset.yaml b/charts/superset/templates/statefulset.yaml
index b8007cd..1c31463 100644
--- a/charts/superset/templates/statefulset.yaml
+++ b/charts/superset/templates/statefulset.yaml
@@ -102,3 +102,4 @@ spec:
         resources:
           requests:
             storage: {{ .Values.storageSize }}
+  serviceName: foo
\ No newline at end of file
```

tarrow (approved):
amazing; let's go then!

PR #1661, opened 2024-06-26
chore(tailscale): remove Elasticsearch

AndrewKostka:


tarrow (approved):
We no longer need this so makes perfect sense to remove

PR #1662, opened 2024-07-03
WIP: deploy UI via ArgoCD on local env

deer-wmde:
WIP

## Describe the changes
- **add placeholder values files for argocd-config chart**
we might want to specify something here, not used right now

- **add argocd-config chart**
this chart effectively configures the argo instance with our deployments, such as the app-of-apps Application and the project and repositories for it

- **add argocd-apps chart**
this chart is the app-of-apps chart: it contains the configurations for the different components (right now, only UI)

- **add argo config to helmfile**
helmfile deploys the argo config chart, which enables argo to deploy the app-of-apps chart

- **move ui value files & create symlinks**
while we can reuse our current value files as they are, I moved them into the charts directory, because otherwise we can not access them from within the chart *I think*. I created symlinks instead so helmfile functionality and thus  skaffold still works.

## Link to Phabricator
https://phabricator.wikimedia.org/T360876

### Prior discussion
Too much to sum it up here, but a lot of sparring with tom in the recent weeks


k8s/helmfile/helmfile.yaml
```diff
@@ -104,6 +104,16 @@ releases:
   ################################
   # ALL ENVIRONMENTS
   ################################
+  - name: argocd-config
+    namespace: argocd
+    chart: ../../charts/argocd-config
+    values:
+      - env/{{`{{ .Environment.Name }}`}}/base.yaml
+      - env/{{`{{ .Environment.Name }}`}}/private.yaml
```
deer-wmde, line 112: I found out that private.yaml can actually be merged into base.yaml, but since that refactor touches a lot of files I refrained from doing it here

charts/argocd-config/templates/applications.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: app-of-apps
+spec:
+  destination:
+    server: https://kubernetes.default.svc
+  project: default
+  source:
+    path: charts/argocd-apps
+    repoURL: {{ .Values.repoUrls.main }}
+    targetRevision: de/argo-apps-ui # debug! needs to be set to HEAD
```
deer-wmde, line 12: this needs to change before merging

charts/argocd-apps/templates/wbaas-ui.yaml
```diff
@@ -0,0 +1,54 @@
...
+        namespace: default
+      source:
+        repoURL: {{ .Values.repoUrls.charts }}
+        targetRevision: HEAD
+        path: charts/ui
+        helm:
+          parameters:
+            - name: helm.debug
+              value: "true"
+          values: '{{ toYaml .Values }}'
```
deer-wmde, line 50: this is currently incorrectly parsed by helm instead of argo

deer-wmde:
closing in favor of #1686

PR #1663, opened 2024-07-03
refactor: argo password reset script

deer-wmde:
some improvements to the password reset script, so that per default its only output is the new password.

### DIFF
--- a/bin/reset-argocd-password
+++ b/bin/reset-argocd-password
@@ -1,17 +1,29 @@
 #!/bin/bash
 
+function patch_password() {
+	ENCODED_HASH="$1"
+
+	kubectl -n argocd patch secret argocd-secret \
+	  --type='json' \
+	  --patch="[{'op': 'replace', 'path': '/data/admin.password', 'value': '${ENCODED_HASH}'}]"
+}
+
 PASSWORD=$(pwgen -N1 -Bs 32)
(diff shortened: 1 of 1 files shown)


bin/reset-argocd-password
```diff
@@ -1,17 +1,29 @@
...
 
+function patch_password() {
+	ENCODED_HASH="$1"
+
+	kubectl -n argocd patch secret argocd-secret \
+	  --type='json' \
+	  --patch="[{'op': 'replace', 'path': '/data/admin.password', 'value': '${ENCODED_HASH}'}]"
+}
+
 PASSWORD=$(pwgen -N1 -Bs 32)
```
m90, line 11: Unrelated to this change, but is pwgen a commonly installed tool? I've never heard of it yet, and it's not installed on my system. Maybe we can get away with reading from `urandom` which should work kind of everywhere

```
tr -dc A-Za-z0-9 </dev/urandom | head -c 13
```

Feel free to disregard this.

deer-wmde: I think it's [fairly common](https://manpages.ubuntu.com/manpages/noble/man1/pwgen.1.html) but I still like your suggestion, thanks :sparkler:

m90 (approved):

PR #1664, opened 2024-07-04
chore(minikube): increase k8s version to 1.28.9

tarrow:
## Describe the changes
This increases the k8s version to match the GKE one in production.

Shouldn't need any more precise review (obviously this won't start a new cluster until make `minikube-start` is called on each engineer's machine.

### Prior discussion
As discussed in the daily this means it's probably also a good point for engineers to remember to update their `kubectl` client

Bug: T369255

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ help:
 .PHONY: minikube-start
 minikube-start: # @HELP Start a local k8s cluster using minikube
 minikube-start:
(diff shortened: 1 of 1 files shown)


deer-wmde (approved):
bonus points if you update .tool-versions as well https://github.com/wmde/wbaas-deploy/blob/main/.tool-versions#L1

PR #1667, opened 2024-07-04
chore(superset): fix volume permissions on staging

AndrewKostka:


charts/superset/values.yaml
```diff
@@ -47,7 +47,9 @@ serviceAnnotations: {}
 
 podAnnotations: {}
 
-securityContext: {}
+securityContext:
+  fsGroup: 1000
```
tarrow, line 51: Could you explain why this is set to 1000? I was trying to get my head around the security context stuff but from what I was reading from the docs. I'mm guessing you tested this with your dibs on thursday and it worked but I was hoping I could figure out why and I couldn't quite explain it myself

(https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) it reads to me like fsGroup is "normally" set to `2000` and that we also ought to the set `runAsUser` and `runAsGroup`

tarrow (approved):

tarrow:
We had a chat about this and we concluded that in this image was actually running as root and then downgrading to `superset` which is a member of group id `1000` which is why this works

PR #1669, opened 2024-07-08
feat(superset): enable lightweight version on production

AndrewKostka:


tarrow (approved):
lgtm; I assume the "creating this user" thing also needs to be manually run

tarrow:
We talked about this in person: specifically we were talking about creating the mariadb observer user so that when the release is installed in production it can start and access the cluster

PR #1681, opened 2024-07-25
chore(make): use kubectl port-forward for local cluster traffic

AndrewKostka:


Makefile
```diff
@@ -39,9 +39,9 @@ minikube-dashboard:
 	minikube --profile minikube-wbaas dashboard
 
 .PHONY: minikube-tunnel
-minikube-tunnel: # @HELP Open a tunnel to the local cluster and expose it on an IP on the host system
+minikube-tunnel: # @HELP Open a tunnel to the local ingress via kubectl and expose it at port 80 on the host system
```
tarrow, line 42: maybe it would make sense to introduce a new target to be called like `make minikube-port-forward` or something?

Makefile
```diff
@@ -39,9 +39,9 @@ minikube-dashboard:
...
 minikube-tunnel:
-	./bin/minikube-tunnel
+	sudo -E kubectl --context minikube-wbaas port-forward -n kube-system service/ingress-nginx-controller 80:80
```
tarrow, line 44: unfortunately this doesn't work for me: I get `error: context "minikube-wbaas" does not exist` since kubectl run with sudo doesn't share contexts with my user

tarrow: `sudo kubectl config get-contexts`  returns no contexts

tarrow (approved):

PR #1689, opened 2024-07-30
(argo) add makefile targets

deer-wmde:
## Describe the changes
Adds some makefile targets to sync argo applications (ui and all, via app-of-apps)
and to list available argo applications

This uses a simple bash script `./bin/argocli` that executes the argocd cli client in the application controller pod. This can also be used on its own to use the cli tool in a generic way.

## Link to Phabricator
https://phabricator.wikimedia.org/T360876

### Prior discussion
https://phabricator.wikimedia.org/T360876#10001045


dati18 (approved):
look great to me

PR #1731, opened 2024-08-19
generate-values: don't use relative path for helmfile

tarrow:
## Describe the changes
We think that this fails when running from a place other than the root of this repo.

We saw it error with `error: helmfile not found: 'k8s/helmfile/helmfile.yaml'` when trying to call this from and action in the ui repo.

### Prior discussion
Pair programmed with @deer-wmde


tarrow:
Abandoned because this doesn't actually work well using a helmfile in docker type situation

PR #1738, opened 2024-08-20
gh-action: try using official yq action

tarrow:
## Describe the changes

Rather than using 3rd party action let's try upstream one. We are unsure if
this will work to install `yq` or if we need to pass it with a `with: cmd:` thing.


tarrow:
supersceeded by #1739

PR #1739, opened 2024-08-20
gh-action: use yq as a builtin

tarrow:
## Describe the changes
See if yq is actually just a builtin


deer-wmde (approved):

tarrow:
FWIW we confirmed this by looking in https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

PR #1756, opened 2024-08-23
argo: anonymous read-only access

deer-wmde:
In order to make wbcloud lives easier I configured a read-only access for ArgoCD where you do not need to log in in order to inspect argo resources with its ui.

I'm not sure I did this right, but it seems to do the trick. Here are some docs: https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/rbac.md

### DIFF
--- a/k8s/helmfile/env/production/argo-cd-base.values.yaml.gotmpl
+++ b/k8s/helmfile/env/production/argo-cd-base.values.yaml.gotmpl
@@ -8,6 +8,7 @@ dex:
 
 configs:
   cm:
+    users.anonymous.enabled: true
     admin.enabled: false
 
     ui.bannerpermanent: true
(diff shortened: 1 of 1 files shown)


tarrow:
This would still be intended only for "developer" use right? Rather than on the public internet?

I would actually see something nice about anoyone being able to see argo rolling live but I'm worried about things like secrets accidentally being exposed since Argo is very highly privileged

deer-wmde:
As of right now it's just for "us" as staging & prod do not have ingresses and you need port-forwarding. I think it would be cool to have it public (also easier to access for us) but I share the same concerns as ArgoCDs Web Client might be (or become) exploitable.

tarrow (approved):
cool, let's try i guess then

PR #1757, opened 2024-08-26
local new-cluster script: add command for running initial migrations

deer-wmde:
Sometimes when initializing a new local cluster it can happen that the artisan pre-install job fails. We should probably fix the Backoff limit of that job or something but for now I'll put this here to save us some time, because that's what this script is there for

### DIFF
--- a/bin/local/new-local-cluster.sh
+++ b/bin/local/new-local-cluster.sh
@@ -38,6 +38,10 @@ set -e
 sleep 60
 helmfile --environment local sync
 
+# run initial database migrations in case the pre-install-job failed for some reason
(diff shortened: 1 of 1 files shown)


tarrow (approved):
Looks find to me but is clearly a hack; I'd probably have favoured something less of a hack in the following order:

1. figuring out why this doesn't actually happen in time anyway
2. increasing the retry count or something to make it more resilient
3. simply recreating the job from the existing job (`k get job api-pre-install-artisan-setup -o json | yq <stuff> | k replace --force -f -`)
4. running this exact hardcoded command but in a new pod/job rather than abusing the existing one for the backend-deployment

deer-wmde:
appreciate the feedback. I'm also curious what the reason for this is, but for now I choose not to follow down that rabbit hole

PR #1765, opened 2024-08-28
Bump table_open_cache and table_definition_cache

dati18:
## Describe the changes

We bumped `table_open_cache` and `table_definition_cache`. Reasons:
- We think the `PlatformStatsSummaryJob` from the `api` opens many tables. The more wikis, the more tables are opened so as the wikis number grows, we need more room for tables to be opened at the same time.
- We think this number should be >3 times the number of wikis (incl. deleted wikis).
- So we should now have around 300 wikis headroom.

- Just try and see if it fixes the problem

### Prior discussion

- Paired with @tarrow and @deer-wmde on this topic.

Bug: T373448

### DIFF
--- a/k8s/helmfile/env/local/sql.values.yaml.gotmpl
+++ b/k8s/helmfile/env/local/sql.values.yaml.gotmpl
@@ -44,8 +44,8 @@ primary:
     max_binlog_size=107374182
 
     # T348842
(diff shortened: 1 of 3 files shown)


tarrow (approved):
right, let's try it

PR #1770, opened 2024-09-16
Remove argocd-apps

dati18:
## Describe the changes

This removes the argocd-apps chart from the repo. This can be merged after the places using this chart have been refactored.

## Link to Phabricator
Bug: T374418

### Prior discussion
This was done in a pairing session with @tarrow


tarrow (approved):
We now think that everything has moved

PR #1780, opened 2024-09-20
staging: platform-nginx route PMs wiki to alpha mediawiki

tarrow:
## Describe the changes
This change routes all wiki traffic that goes to the platform-nginx for https://anton2.wikibase.dev/ to the alpha mediawiki deployment.

This allows us to push specific images there for unusual/dangerous configurations.

Right now there is are no values change in git for this unusual configuration and they are being managed by manually altering the images that are present.

Currently being tried under "dibs" on staging. You could check that this is indeed working as expected by confirming that traffic is going to the alpha wiki.

### Prior discussion
Almost none, let me know what you think @rosalieper 

Bug: T374336

### DIFF
--- a/k8s/helmfile/env/staging/platform-nginx.nginx.conf
+++ b/k8s/helmfile/env/staging/platform-nginx.nginx.conf
@@ -70,6 +70,11 @@ server {
                 set $mwgroup "alpha";
         }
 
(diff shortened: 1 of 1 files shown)


rosalieper (approved):

rosalieper:
Thank you for doing this!! Really nice hack

PR #1785, opened 2024-09-23
argocd: fix environment banner

deer-wmde:
A while ago I configured different banners for the different environment argocd runs in. For some reason it stopped working some time, now I found the reason (and I fixed the URL for local env)

image
image
image


dati18 (approved):
nice change.

dati18 (approved):

PR #1788, opened 2024-09-27
set cpu limits to 0 for relevant workloads

deer-wmde:
## Describe the changes
- set cpu limits to 0 for most workloads for production

- does it make sense leaving out certain workloads like argo and metric stuff?
- is it a good idea to do this on production first?
  - not sure if we are able to "test" it properly on staging

## Link to Phabricator
https://phabricator.wikimedia.org/T375874

### Prior discussion
with @tarrow


rosalieper (approved):
LGTM!

tarrow (approved):

PR #1790, opened 2024-09-30
lift cpu limits for mediawiki resources

deer-wmde:
https://phabricator.wikimedia.org/T375874

as a first step we only lift cpu limits for mw resources and will observe the effect

### DIFF
--- a/k8s/helmfile/env/production/mediawiki-139.values.yaml.gotmpl
+++ b/k8s/helmfile/env/production/mediawiki-139.values.yaml.gotmpl
@@ -59,26 +59,26 @@ resources:
       cpu: 500m
       memory: 1600Mi
     limits:
-      cpu: 0
+      cpu: null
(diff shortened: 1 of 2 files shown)


dati18 (commented):
In my opinion, we should bump more on `production` and less on `staging` since `staging` is not as demanding as `prod`. What do you think?

tarrow:
Generally looks fine to me; I struggled to find some docs saying that 0 === unlimited. I don't suppose you have a reference for that?

deer-wmde:
> Generally looks fine to me; I struggled to find some docs saying that 0 === unlimited. I don't suppose you have a reference for that?

Excellent question, I will verify this on the staging cluster.

tarrow (approved):

tarrow:
given you're not here I'm gonna ship this now :)

PR #1793, opened 2024-10-01
chore(tf): add compute-pool-2 on staging

AndrewKostka:
This commit adds a new node pool to replace the exiting one.
It uses smaller machines to free up some capacity.
The commit removing the existing node pool will happen later.

Bug: T376117


tarrow (approved):
lgtm; I added more content + the bug to the PR; can you also add it when you squash to the commit message?

PR #1798, opened 2024-10-07
chore(tf): update Google provider to 6.5.0 on staging

AndrewKostka:
Bug: T376141


tarrow (approved):
I'm interested that this is all it needs to work; I would have sort of expected some `location` to be needed on the cluster but if this works fantastic.

Properly understanding this is a little opaque to me still

PR #1800, opened 2024-10-10
Local/Staging: Use api 0.32.0

rosalieper:
Update api  chart version for staging and local

## Link to Phabricator
Bug: T375268

### DIFF
--- a/k8s/helmfile/helmfile.yaml
+++ b/k8s/helmfile/helmfile.yaml
@@ -163,7 +163,7 @@ releases:
 
   - name: api
     chart: wbstack/api
-    version: "0.31.2"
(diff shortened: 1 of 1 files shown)


k8s/helmfile/helmfile.yaml
```diff
@@ -163,7 +163,7 @@ releases:
 
   - name: api
     chart: wbstack/api
-    version: "0.31.2"
+    version: '{{- if eq .Environment.Name "production" }} "0.31.2" {{else}} "0.32.0" {{end}}'
```
tarrow, line 166: `in ./helmfile.yaml: [improper constraint:  "0.32.0" ]` seems to be what I see if I try to `diff` this to staging. Do you see that?

rosalieper: yes I see that.

PR #1821, opened 2024-10-29
queryservice: remove transitional allowlist values

AndrewKostka:
Bug: [T375384](https://phabricator.wikimedia.org/T375384)


tarrow (approved):
I think this is find but we'll need to make sure that the configmap is repopulated (by running the api job) right after deploying because I believe merging this patch will blank the configmap

PR #1833, opened 2024-11-06
docs: add some content for deploying with argo

tarrow:


deer-wmde:
Thanks for adding this!

as mentioned elsewhere, I'm currently unclear what is the right way to treat argo controlled deployments in the helmfile (to `installed: false` or not to? see https://github.com/wmde/wbaas-deploy/pull/1781) so we should probably verify this one more time

tarrow:
So I think `installed: false` is not right but also leaving it there isn't right either. I think a good solution would be what is shown here in #1840

PR #1837, opened 2024-11-12
Preserve container images when recreating minikube

AndrewKostka:


deer-wmde:
If I remember correctly this worked pretty good, did we observe any negative side effects so far? otherwise I'd be happy with merging this

PR #1847, opened 2024-11-15
helmfile: fixup for argocd-config 1.0.8

tarrow:
Unfortunately 1.0.8 was mistakenly configured to use https://github.com/wbstack/charts.git rather than
https://wbstack.github.io/charts and therefore wasn't correctly parsing the chart without a path. 1.0.9 is a fixup for this that uses the correct repoURL

Bug: T375195

### DIFF
--- a/k8s/helmfile/helmfile.yaml
+++ b/k8s/helmfile/helmfile.yaml
@@ -109,7 +109,7 @@ releases:
   - name: argocd-config
     namespace: argocd
     chart: wbstack/argocd-config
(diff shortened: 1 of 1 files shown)


dati18 (approved):
I approve, but create a PR for 1.0.9 for all envs so we can move it as soon as this one works

tarrow:
Good idea! I already updated https://github.com/wmde/wbaas-deploy/pull/1815

PR #1848, opened 2024-11-15
staging(argocd): increase memory request

tarrow:
We seem to regularly be being OOM killed so this commit increases the request to match the limit of 512Mi

Bug: T375195


dati18 (approved):
I hove 512 is enough

PR #1864, opened 2024-11-20
argocd: add redis-2 on staging

AndrewKostka:


tarrow (approved):
obviously to ship post chart merges

PR #1868, opened 2024-11-22
enable TLS for local environment

deer-wmde:
> [!NOTE]
> depends on: https://github.com/wbstack/charts/pull/179

## Describe the changes
- changes local tunnel script to serve port `443` instead of `80`
- enables TLS for local ingress
- adds certificate config for self-signed cert-issuer

- I noticed for all environments all cluster issuers get deployed. This is not necessarily an issue (since different environment will have different keys generated) but I'd like to hear if someone is worried about this in any way
- do the api pods need the self-signed cert? I'm not sure
- are there missing hardcoded instances of `http://` that still need to be changed somewhere?
- do you also experience this bug? https://mattermost.wikimedia.de/swe/channels/wikibase-cloud-engineering/x3x9x6nuhtg1pfxwew8wrtjn9w
- any ideas on this? I couldn't figure it out https://github.com/wmde/wbaas-deploy/pull/1868/files#diff-f64d247983d17f7ce876d21cf69063b9f25d737b15d517cf7c72ddf1bdc64c19R86

## Link to Phabricator

### DIFF
--- a/bin/minikube-tunnel
+++ b/bin/minikube-tunnel
@@ -4,7 +4,7 @@ EXTERNAL_IP=$(minikube --profile minikube-wbaas kubectl -- -n kube-system get se
 
 trap 'kill -15 $kids; exit 143' EXIT HUP INT TERM
 
-sudo socat tcp-listen:80,reuseaddr,fork tcp:"$EXTERNAL_IP":80 &
+sudo socat tcp-listen:443,reuseaddr,fork tcp:"$EXTERNAL_IP":443 &
 kids=$!
 
 minikube --profile minikube-wbaas tunnel &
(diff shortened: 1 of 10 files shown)


tarrow:
maybe you already did locally and didn't commit but I wondered if you needed to also touch https://github.com/wmde/wbaas-deploy/blob/main/bin/minikube-tunnel#L7 and open 443 there?

deer-wmde:
> maybe you already did locally and didn't commit but I wondered if you needed to also touch https://github.com/wmde/wbaas-deploy/blob/main/bin/minikube-tunnel#L7 and open 443 there?

that was it, thanks.

tarrow (approved):
Looks good to me; didn't try locally though! Don't forget to merge the chart first! :)

PR #1869, opened 2024-11-25
redis: reduce CPU requests on staging

AndrewKostka:


tarrow (approved):

tarrow:
Looks like this "fails" due to an oversight in removing cpu limits for the master as well as the replica. See: #1871

PR #1872, opened 2024-11-27
Staging: Route 2 wikis to alpha mediawiki deployment

rosalieper:
Bug: T378743


tarrow (approved):
Looks good! I'm not sure how best to test this locally so I think it would be better to just try it on staging!

I really like how you added a separate section and properly documented what that section is for

PR #1877, opened 2024-11-27
cleanup private.yaml (staging/prod)

deer-wmde:
## Describe the changes
While looking at local TLS setup and trying to understand our certificate setup I came across some things that can be cleaned up.

- removed certificate config from private.yaml
  - as this seems unused and moved to certificates.values.yaml.gotmpl in the past
- removed a confusing comment, because as far as I can tell that value is still used on the other hand

Diffs should result in *no changes* for staging or production

## Link to Phabricator


tarrow (approved):
I believe after checking that this is (as you say) a noop on prod +staging

PR #1883, opened 2024-12-02
remove argo cpu limits

deer-wmde:
There was some discussion recently about this, but I fail to remember the outcome why this was added and then reverted and such. All I know is that without this currently a local cluster re-initialization is not working, so we should change *something*. Change requests/feedback/etc welcome


tarrow:
I'm not sure why it was reverted; let's merge it!

tarrow (approved):

PR #1897, opened 2024-12-11
argo: disable dex in all environments

tarrow:
this commit removes the resources setting that appears to be inadvertantly re-enabling the dex server and config after it was attempted to be disabled at the top of
production/argo-cd-base.values.yaml.gotmpl

Bug: T381960


deer-wmde (approved):

tarrow:
so I'll just ship this instead of #1896

PR #1898, opened 2024-12-11
enable wikibasequalitycontraint on wbqc-2-alpha.wikibase.dev

rosalieper:
Bug: T378743


tarrow:
some cheeky self-merging ;) ?

rosalieper:
since it was the 3rd wiki I was creating for this task plus it was a just a wiki name change, I wanted to deploy it to alpha quickly and let Anton try the wiki. But truth is I should have ask another eye still, Sorry!

PR #1901, opened 2024-12-12
prod: use argocd-config 1.1.0

deer-wmde:
https://phabricator.wikimedia.org/T377585

https://github.com/wmde/wbaas-deploy/pull/1885/files

uses argocd-config chart v0.1.0 on production


tarrow (changes requested):
Unless https://github.com/wmde/wbaas-deploy/pull/1910/files#r1883786844 is done this needs that change here

tarrow (approved):

PR #1903, opened 2024-12-12
local tls: add ca cert export and documentation + argo cert

deer-wmde:

This adds a makefile target to easily export the local CA certiface that gets used to self-sign the locally used TLS certificates. This can get used to circumvent browser warnings.

- also adds config to issue local argo a certificate

### DIFF
--- a/doc/local-dev-env.md
+++ b/doc/local-dev-env.md
@@ -245,6 +245,15 @@ sudo wget https://raw.githubusercontent.com/roboll/helmfile/master/autocomplete/
 sudo sh -c 'skaffold completion bash > /usr/share/bash-completion/completions/skaffold'
 ```
 
+## [Optional] install local CA certificate
+Since we [introduced](https://phabricator.wikimedia.org/T378691) using HTTPS for local ingresses, you will get a scary warning when accessing local web interfaces. This can be mitigated by trusting the local CA certificate that is getting used for self-signing. The easiest way to do this is to save the local CA certificate in a file by accessing the secret it lives in (`wikibase-local-tls`) and importing it in your browser. There is also the possibility to import it into the trust store of your operating system, for example via the tool [mkcert](https://github.com/FiloSottile/mkcert), but you should be aware of the possible consequences this could have for the security of your machine.
+
+> [!TIP]
+> Running `make local-ca` will save the certificate to the file `wikibase-local-tls.crt`. It is highly recommended to delete the file again after importing it.
+
+> [!NOTE]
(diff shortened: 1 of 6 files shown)


tarrow (approved):

deer-wmde:
I'm super confused right now - I thought I was able to import the certificate into firefox yesterday but now it doesnt seem to work?!

deer-wmde:
for some reason the CA cert in the secret is not a CA cert but the same cert as the one in `data.tls.crt` - I have no idea if this changed now since yesterday somehow but I'm 99% sure I was able to import it into firefox, but I also have no idea what could've changed now.

deer-wmde:
I can only imagine I made a mistake yesterday. But I found a way to do this now https://cert-manager.io/docs/configuration/selfsigned/#bootstrapping-ca-issuers

This will require a chart update though.

deer-wmde:
adjusted this accordingly and the new chart PR is here: https://github.com/wbstack/charts/pull/184

deer-wmde:
sorry, this should be it's own PR but now slipped in. I was curious how hard it would be to get argocd now locally a valid cert, and it required only very few changes, so I added it (requires one more chart change though to set the namespace https://github.com/wbstack/charts/pull/185)

tarrow (approved):
very pro

doc/local-dev-env.md
```diff
@@ -245,6 +245,15 @@ sudo wget https://raw.githubusercontent.com/roboll/helmfile/master/autocomplete/
 sudo sh -c 'skaffold completion bash > /usr/share/bash-completion/completions/skaffold'
 ```
 
+## [Optional] install local CA certificate
+Since we [introduced](https://phabricator.wikimedia.org/T378691) using HTTPS for local ingresses, you will get a scary warning when accessing local web interfaces. This can be mitigated by trusting the local CA certificate that is getting used for self-signing. The easiest way to do this is to save the local CA certificate in a file by accessing the secret it lives in (`wikibase-local-tls`) and importing it in your browser settings. There is also the possibility to import it into the trust store of your operating system, for example via the tool [mkcert](https://github.com/FiloSottile/mkcert), but you should be aware of the possible consequences this could have for the security of your machine.
+
+> [!TIP]
+> Running `make local-ca` will save the certificate to the file `wikibase-local-tls.crt`. It is highly recommended to delete the file again after importing it.
```
tarrow, line 252:
```suggestion
> Running `make local-ca` will save the certificate to the file `wikibase-local-tls.crt`.
```

If I understand correctly a certificate isn't something secret we need to protect. Deleting it would be convenient for not getting mixed up but it's not the tls.key which I *think* is what you want to keep secret

PR #1910, opened 2024-12-13
local/staging: use argocd-config 1.1.0

deer-wmde:
attempt #3 to ship https://github.com/wmde/wbaas-deploy/pull/1885

### DIFF
--- a/k8s/helmfile/env/production/argocd-config.values.yaml.gotmpl
+++ b/k8s/helmfile/env/production/argocd-config.values.yaml.gotmpl
@@ -1,2 +1,2 @@
 environment: production
-
+appOfAppsVersion: 1.1.0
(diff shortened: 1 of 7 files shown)


k8s/helmfile/env/production/argocd-config.values.yaml.gotmpl
```diff
@@ -1,2 +1,2 @@
 environment: production
-
+appOfAppsVersion: 1.1.0
```
tarrow, line 2: maybe set this to 1.1.1 anyway since I think it's unused until we for to `argocd-config` `1.1.0` anyway? Otherwise I bet we forget?

tarrow (approved):

k8s/helmfile/env/production/argocd-config.values.yaml.gotmpl
deer-wmde: valid concern but due to coordination difficulties I will move on now to see if this even finally works as intended

PR #1915, opened 2024-12-18
nginx: add robots.txt for Wikibase sites

AndrewKostka:
A slightly reduced version of https://www.wikidata.org/robots.txt


tarrow (approved):
Looks good to me; tested locally and got a robots.txt where I would expect.

I also compared the list to the wikidata one. I wondered for is bit if we should still be including all those non-english instances of `/wiki/Special:` (e.g. `/wiki/Spezial:`) but with some very rough testing it appears that since we enforce an english "site content language" this isn't an issue.

So, let's go and ship to prod

PR #1924, opened 2025-01-29
Prepare for automated deploy commits for tools

AndrewKostka:


k8s/helmfile/env/local/tool-widar.values.yaml.gotmpl
```diff
@@ -1,3 +1,5 @@
-# the local service matches production, so this file is empty
```
tarrow, line 1: what an amusing lie

tarrow (approved):

PR #1987, opened 2025-02-26
docs: add dumping-wikis.md

deer-wmde:
Some documentation on how to create XML/JSON dumps for user wikis in light of https://phabricator.wikimedia.org/T385154


tarrow (approved):
truely awesome; thanks for getting on and also so nicely documenting this

PR #1994, opened 2025-03-06
replace `wbaas.localhost` with `wbaas.dev`

deer-wmde:
https://phabricator.wikimedia.org/T383335#10591561

- mostly just replaces occurences of `wbaas.localhost` with `wbaas.dev`
- updates some documentation


tarrow (approved):
LGTM - I'll be very happy if this applies and everything still works without a little more poking but let's see

PR #2046, opened 2025-04-03
use ingress-nginx 4.12.1 with less strict config

deer-wmde:
use the newer ingress-nginx chart again (compare https://github.com/wmde/wbaas-deploy/pull/2044)

and disable strict path types to keep api and ui ingress path config working

### DIFF
--- a/k8s/helmfile/env/production/ingress-nginx.values.yaml.gotmpl
+++ b/k8s/helmfile/env/production/ingress-nginx.values.yaml.gotmpl
@@ -19,6 +19,7 @@ controller:
     limit-conn-status-code: 429
     custom-http-errors: "502,503"
     preserve-trailing-slash: true
+    strict-validate-path-type: false
     block-user-agents: ~*Seekport\sCrawler.*,~*MJ12bot.*,~Bytespider,~thesis-research-bot
     # https://stackoverflow.com/a/37877244
     log-format-upstream: >-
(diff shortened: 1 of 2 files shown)


deer-wmde:
currently trying to see if we could keep this enabled and change the ingress paths instead

deer-wmde:
So changing the path like this (without wildcard regex) for the UI worked for me, but not for the API.


`/wbaas-deploy/k8s/helmfile/env/production/ui.values.yaml.gotmpl`
```yaml
ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.class: nginx
    nginx.ingress.kubernetes.io/use-regex: "true"
    nginx.ingress.kubernetes.io/from-to-www-redirect: "true"
  hosts:
  - host: {{ .Values.uiHostName }}
    paths:
    - /
```

`/wbaas-deploy/k8s/helmfile/env/local/private.yaml`
```yaml
    ingressHosts:
      - host: api.wbaas.dev
        paths:
          - /
      - host: www.wbaas.dev
        paths:
          - /api/
```

tarrow (approved):
looks good to me; thanks for confirming its only needed for one service

deer-wmde:
> looks good to me; thanks for confirming its only needed for one service

ah sorry I was unclear; the change is needed for both the UI and API ingresses, I just tried to get around disabling that strict setting and altering the ingresses conf instead, which only worked fine for the UI, but not for he API - hence still needing the change merged by this PR

PR #2047, opened 2025-04-03
increase wbaas.dev DNS TTL

deer-wmde:
For me from time to time my system has trouble resolving wbaas.dev - maybe increased TTL will help? Did anyone else experience this?

### DIFF
--- a/tf/env/staging/dns-local.tf
+++ b/tf/env/staging/dns-local.tf
@@ -34,7 +34,7 @@ resource "google_dns_record_set" "local-A" {
   rrdatas = [
     "127.0.0.1"
   ]
-  ttl  = 300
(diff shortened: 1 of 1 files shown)


tarrow:
did not have this problem but... which system struggled to resolve it? Your local firefox? Did you see anything useful from `host`/`dig` etc?

deer-wmde:
yeah I thought firefox at first but its OS wide, so even with curl etc

if no one else sees this I guess its my LAN DNS or something

tarrow (approved):
fair enough; I see no reason to not have a 6hr vs 5m ttl though. Surely we have no plans to move this from `127.0.01` in the near term anyway.

PR #2050, opened 2025-04-07
sql statefulsets: remove cpu limits

deer-wmde:
related: https://phabricator.wikimedia.org/T390823

Currently SQL StatefulSets can not come back up due to invalid cpu limits

```
  Warning  FailedCreate  2m6s (x30 over 3h27m)  statefulset-controller  create Pod sql-mariadb-primary-0 in StatefulSet sql-mariadb-primary failed error: Pod "sql-mariadb-primary-0"
 is invalid: spec.containers[0].resources.requests: Invalid value: "750m": must be less than or equal to cpu limit of 0
```

### DIFF
--- a/k8s/helmfile/env/production/sql.values.yaml.gotmpl
+++ b/k8s/helmfile/env/production/sql.values.yaml.gotmpl
@@ -10,7 +10,6 @@ primary:
       cpu: '2'
       memory: 8Gi
     limits:
-      cpu: null
       memory: 8Gi
   persistence:
     enabled: true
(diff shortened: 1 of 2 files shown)


tarrow (approved):

tarrow:
I guess there are probably more of these null limits out there waiting to bite us

tarrow:
Yes - I already tried to clean this up and then had to revert it. I believe related to the helmfile->argo values issues. We could probably just try to revert the revert again. See: #1842

deer-wmde:
ah.. well since you approved I merged it for now in order to make #2048 work (which I merged before and then was wondering why the workloads are down again :P)  but yeah I think we're now set up to get this same surprise for the  statefulsets as well

PR #2057, opened 2025-04-09
remove cpu limits follow up 

deer-wmde:
follow up to https://github.com/wmde/wbaas-deploy/pull/2050

These are helmfile values only, so should be valid and mostly no-op (apart from staging superset because that value was not `null`). As far as I can see there is one limit left for the argo deployed redis but maybe we should touch that outside of this PR, as I'm currently not sure what brought us this ping-ponging back and forth between removing them and reverting again.


tarrow:
I looked a bit today and got tempted to clean some up; I think we definitely have issues with `queryservice` for example (https://github.com/wbstack/charts/blob/main/charts/queryservice/values.yaml#L46)

PR #2067, opened 2025-04-10
staging: reduce sql cpu requests

deer-wmde:
follow up to https://github.com/wmde/wbaas-deploy/pull/2065

because now ES doesnt get its piece of the pie: `Warning  FailedScheduling   2m7s (x237 over 19h)   default-scheduler   0/3 nodes are available: 1 Insufficient cpu, 2 Insufficient memory. preemption: 0/3 nodes are available: 3 No preemption victims found for incoming pod.`

### DIFF
--- a/k8s/helmfile/env/staging/sql.values.yaml.gotmpl
+++ b/k8s/helmfile/env/staging/sql.values.yaml.gotmpl
@@ -1,7 +1,7 @@
 primary:
   resources:
     requests:
-      cpu: 750m
+      cpu: 350m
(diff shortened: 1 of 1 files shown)


tarrow (approved):
This is fine but you could even drop this to `10m` which would make a lot more space

deer-wmde:
> This is fine but you could even drop this to `10m` which would make a lot more space

yeah it just sounds so little :D but the graphs agree, and worst case it can still eat more

PR #2068, opened 2025-04-10
staging: reduce argo resource requests

deer-wmde:
reduction of argou resource requests on staging

### DIFF
--- a/k8s/helmfile/env/staging/argo-cd-base.values.yaml.gotmpl
+++ b/k8s/helmfile/env/staging/argo-cd-base.values.yaml.gotmpl
@@ -8,10 +8,18 @@ dex:
 controller:
   resources:
    limits:
-     memory: 1024Mi
+     memory: 512Mi
    requests:
      cpu: 250m
-     memory: 1024Mi
+     memory: 256Mi
+
+notifications:
+  resources:
+    limits:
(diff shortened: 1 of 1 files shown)


k8s/helmfile/env/staging/argo-cd-base.values.yaml.gotmpl
```diff
@@ -8,10 +8,18 @@ dex:
 controller:
   resources:
    limits:
-     memory: 1024Mi
+     memory: 512Mi
    requests:
      cpu: 250m
-     memory: 1024Mi
+     memory: 256Mi
```
tarrow, line 14: This is now actually lower than the average usage. May not be an issue but something to keep an eye on if this ends up regularly being evicted at time of high memory usage across the whole cluster

tarrow (approved):

k8s/helmfile/env/staging/argo-cd-base.values.yaml.gotmpl
deer-wmde: Ah interesting, you're right! I just realized the staging requests were much higher than in prod and copied those. But yeah, on prod the usage is also above average, guess I'll adjust it then 
image

PR #2078, opened 2025-04-10
prod(tf): remove old node pool

tarrow:
Follow up to remove this old node pool that still has high throughput logging enabled

Bug: T390698


tarrow:
fixing forward some messing git merge that didn't work right

PR #2097, opened 2025-04-17
prod(k8s): reduce tailscale operator CPU request

tarrow:
Removes cpu limit too in line with other workloads

Bug: T390698


tarrow:
Tested on staging under dibs; this is indeed the correct config to trickle down to the operator replicaset

dati18 (approved):

PR #2098, opened 2025-04-17
prod(k8s): reduce queryservice cpu requests

tarrow:
Values determined by looking at last 2 weeks resource utilisation data and using the new GKE recommended request metric

Bug: T390698


tarrow:
removed some additional stuff about `null` cpu limits I'd like us to do later

dati18 (approved):

PR #2106, opened 2025-04-28
staging/local: use api image 10x.19.0

deer-wmde:
somehow the auto PR creation did not happen for https://github.com/wbstack/api/pull/924


tarrow:
I guess the takeaway from this is that is the auto-PR isn't made there is a high chance that also the image isn't made correctly

PR #2149, opened 2025-06-02
Add solution for "minikube docker context warning" to FAQ

outdooracorn:


tarrow (approved):
Excellent!  I never saw this before; I guess it's somehow an interplay with docker from mwcli and contexts that we don't regularly run into

addshore:
> I guess it's somehow an interplay with docker from mwcli and contexts that we don't regularly run into

Yeh, mwcli doesnt make use of any sort of docker context functionality. It's "contexts" are its own thing.

Interestingly, I also have 2 contexts out the gate

```
docker context ls
NAME            DESCRIPTION                               DOCKER ENDPOINT                             ERROR
default *       Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
desktop-linux   Docker Desktop                            npipe:////./pipe/dockerDesktopLinuxEngine
```

PR #2150, opened 2025-06-02
Fix local terraform dependencies

outdooracorn:
During initial setup of development environment, `tofu apply` resulted in a `Error: namespaces "adhoc-jobs" not found`. Given that `resource "kubernetes_namespace" "adhoc-job-namespace"` exists, I assumed this was simply because resources tried to use the "adhoc-jobs" namespace before it had been created. Explicitly defining the dependency resolved the issue.

### DIFF
--- a/tf/env/local/secrets-recaptcha.tf
+++ b/tf/env/local/secrets-recaptcha.tf
@@ -1,5 +1,9 @@
 resource "kubernetes_secret" "recaptcha-v3-secrets" {
-  for_each = toset(["default", "api-jobs", "adhoc-jobs"])
+  for_each = toset([
+    "default",
+    kubernetes_namespace.api-job-namespace.metadata[0].name,
+    kubernetes_namespace.adhoc-job-namespace.metadata[0].name,
+  ])
   metadata {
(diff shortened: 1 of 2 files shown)


tf/env/local/secrets-recaptcha.tf
```diff
@@ -29,6 +34,11 @@ resource "kubernetes_secret" "recaptcha-v2-secrets" {
     "site_key"   = var.recaptcha_v2_site_key,
     "secret_key" = var.recaptcha_v2_secret
   }
+
+  depends_on = [
```
tarrow, line 38: Good idea to be explicit to avoid the race. I think if it works it would be even better to allow terraform to resolve the dependency itself.

I think this means on line 26 we should remove the raw `"adhoc-jobs"` string and replace it with a reference to a string from the kubernetes namespace resource (`kubernetes_namespace.adhoc-job-namespace.metadata[0].name`)

You can see an example of this in the namespaces.tf: https://github.com/wmde/wbaas-deploy/blob/9571a3c70a779dc2921dac9cb590cec52e100b34/tf/env/local/namespaces.tf#L10

PR #2151, opened 2025-06-04
Update local-dev-env.md

outdooracorn:
* Update doc around the `terraform.tfvars` file and recaptcha keys
* Update doc around `helmfile diff` not being possible before a `helmfile apply` on a new install
* Remove outdated information that `make diff-local` and `make apply-local` commands use the `--skip-deps` argument
* Use github flavoured markdown for notes (`> [!NOTEs]`)
* Updated list of pods expected to be seen
* Fix commands to connect to the 'redis master' container
  * 'redis' has one 'd' in the `kubectl` command
  * Use the correct arguments with the `k8s-shell` command
* Fix some typos

### DIFF
--- a/doc/local-dev-env.md
+++ b/doc/local-dev-env.md
@@ -71,15 +71,15 @@ tofu init
 
 For convenience, you can store local secrets for the cluster in `.tfvars` files which will be ignored by git. 
 
-Create a `terraform.tfvars` file in `tf/env/local` and add the recaptcha secrets to it. You can use [test keys][test-keys] for the v2 secrets. For v3 you will have to create your own secrets via the [v3 admin console](https://www.google.com/recaptcha/admin).
+Create a `terraform.tfvars` file in `tf/env/local` and add the recaptcha secrets to it. You can use [test keys][test-keys] for the v2 secrets. For v3 use the "wbaas.dev v3" keys from the [admin console](https://www.google.com/recaptcha/admin).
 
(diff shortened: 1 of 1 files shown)


doc/local-dev-env.md
```diff
@@ -71,15 +71,15 @@ tofu init
 
 For convenience, you can store local secrets for the cluster in `.tfvars` files which will be ignored by git. 
 
-Create a `terraform.tfvars` file in `tf/env/local` and add the recaptcha secrets to it. You can use [test keys][test-keys] for the v2 secrets. For v3 you will have to create your own secrets via the [v3 admin console](https://www.google.com/recaptcha/admin).
+Create a `terraform.tfvars` file in `tf/env/local` and add the recaptcha secrets to it. You can use [test keys][test-keys] for the v2 secrets. For v3 use the "wbaas.dev v3" keys from the [admin console](https://www.google.com/recaptcha/admin).
```
outdooracorn, line 74: Still need to check if "v2 test keys" can be used, or we should move to using `wbaas.dev v2` keys

deer-wmde: If I remember correctly the platform UI uses v3 and mediawiki the v2 keys - but I'm not sure

outdooracorn: Yes, correct, thanks for the tip!

The [v2 test keys](https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do) still work, and I think they are the best to use as they won't ask you to solve a puzzle.

PR #2171, opened 2025-06-23
Fix missing Bitnami image for Elasticsearch

AndrewKostka:
Bug: T397604


k8s/helmfile/env/production/elasticsearch-2.values.yaml.gotmpl
```diff
@@ -70,3 +70,10 @@ ingest:
   enabled: false
 
 clusterName: {{ .Release.Name }}
+
+volumePermissions:
```
tarrow, line 74: Assuming we're setting this so if we use this image in the future we won't be bitten by the same thing. From what I could see we don't currently use this one.

PR #2191, opened 2025-07-24
Add Alibaba DC ips to very tight rate limit

tarrow:
This add a new limit_req_zone that only applies to ips in a list of CIDRs. This is done by specifying a `geo`[1] and a `map`[2] that work in conjunction to limit all of these requests together rather than treat them as separate ips.

[1] https://nginx.org/en/docs/http/ngx_http_geo_module.html
[2] https://nginx.org/en/docs/http/ngx_http_map_module.html

Bug: T400046


tarrow:
Merged but so far I'm not seeing the log lines I'd expect (e.g. containing `delaying request, excess: <n>, by zone "cloud_provider"`

PR #2206, opened 2025-08-13
Set pathType to Prefix for Anubis ingress

AndrewKostka:
Bug: [T399851](https://phabricator.wikimedia.org/T399851)


outdooracorn (approved):
Thanks! :shipit:

PR #2226, opened 2025-08-28
Use wbstack mirror for Bitnami charts used by Helmfile

AndrewKostka:
Bug: [T401478](https://phabricator.wikimedia.org/T401478)


deer-wmde (approved):
LGTM; kubectl reports no diffs, which is expected I think

PR #2241, opened 2025-09-03
Add script to update existing ingresses of custom domains

AndrewKostka:
Bug: [T399851](https://phabricator.wikimedia.org/T399851)


deer-wmde:
looks solid to me! I think adding a comment to the top of the script for the intended purpose would be grand (switching between platform-nginx and anubis)

PR #2253, opened 2025-09-22
fix: bitnami chart URLs

deer-wmde:
- We noticed the index.yaml file we used here still pointed to artifacts on bitnami servers: https://raw.githubusercontent.com/wbstack/bitnami-legacy/14b6efaa84c7ed5684885ca8d29479fe6e3ed973/bitnami/index.yaml

- We packaged the affected charts and uploaded them as releases here https://github.com/wbstack/bitnami-legacy/releases

- Discussion: https://mattermost.wikimedia.de/swe/pl/smgair3ccpnzfbzqhw665ifyte

Bug: T405468

### DIFF
--- a/k8s/helmfile/helmfile.yaml
+++ b/k8s/helmfile/helmfile.yaml
@@ -1,8 +1,6 @@
 repositories:
   - name: wbstack
     url: https://wbstack.github.io/charts
-  - name: bitnami-legacy
-    url: https://raw.githubusercontent.com/wbstack/bitnami-legacy/14b6efaa84c7ed5684885ca8d29479fe6e3ed973/bitnami
   - name: cetic
     url: https://cetic.github.io/helm-charts
   - name: codecentric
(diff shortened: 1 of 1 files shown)


outdooracorn:
> Sounds like maybe the chart we've packaged up in the legacy repo isn't the same as the chart we were using previously?

Could well be. We obviously tried to use the same chart. One thing we did which I'm not sure about is a `helm dependency update` which might be the cause? Pulled in some dependency that changed slightly?

> Just to confirm when you say:
> 
> > I have noticed that this change adds app.kubernetes.io/version: "8.8.1" as a label.
> 
> When and how do you see that? After applying this commit to a cluster that was in which state? And how are you applying it?

This is by checking out the PR and doing a diff with production (`make diff-production`)

deer-wmde:
> One thing we did which I'm not sure about is a helm dependency update which might be the cause? Pulled in some dependency that changed slightly?

I was also suspecting this but I just built another package without that step and I still see that diff :thinking: https://github.com/wbstack/bitnami-legacy/releases/tag/elasticsearch%2F19.10.2-B

outdooracorn:
Ahh, we should have used `helm dependency build` instead of `helm dependency update` as that will use the existing `Chart.lock` file rather than update it.

You might also need to make sure that you don't have a `charts/` dir before you run `helm dependency build`, but I'm not sure on that.

PR #2696, opened 2026-04-22
Add warning about T423380

tarrow:


k8s/helmfile/env/production/queryservice.values.yaml.gotmpl
```diff
@@ -1,3 +1,4 @@
+# Caution! Changing values here is associated with T423380
```
outdooracorn:
```suggestion
# Caution! Changing values here will blank the `allowlist.txt` key in the `queryservice-allowlist` ConfigMap (T423380)
```

k8s/helmfile/env/production/queryservice.values.yaml.gotmpl
```diff
@@ -1,3 +1,4 @@
+# Caution! Changing values here will blank the `allowlist.txt` key in the `queryservice-allowlist` ConfigMap (T423380)
```
tarrow:
```suggestion
# Caution! Changing values here will silently blank the `allowlist.txt` key in the `queryservice-allowlist` ConfigMap (T423380)
```

tarrow: @outdooracorn ^^

outdooracorn (approved):

PR #2711, opened 2026-05-03
anubis: Allow all HEAD requests

tarrow:
Add an allow rule for all HEAD requests with a method expression as per the docs [1]

[1] https://anubis.techaro.lol/docs/admin/configuration/expressions

Bug: T419979

### DIFF
--- a/k8s/helmfile/env/production/anubis.values.yaml.gotmpl
+++ b/k8s/helmfile/env/production/anubis.values.yaml.gotmpl
@@ -22,8 +22,8 @@ botPolicy:
           - 'path.startsWith("/w/rest.php")'
           - 'path.startsWith("/query/sparql")'
     - name: allow-head-requests
-        action: ALLOW
-        expression: 'method == "HEAD"'
+      action: ALLOW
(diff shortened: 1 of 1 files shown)


outdooracorn (approved):
LGTM!

k8s/helmfile/env/production/anubis.values.yaml.gotmpl
```diff
@@ -21,6 +21,9 @@ botPolicy:
           - 'path.startsWith("/w/api.php")'
           - 'path.startsWith("/w/rest.php")'
           - 'path.startsWith("/query/sparql")'
+    - name: allow-head-requests
```
addshore, line 24: I believe this is invalid YAML
SO I recreated https://github.com/wmde/wbaas-deploy/pull/2714 from a better branch on my fork (as it was on main and auto closed before when I cleaned it up)

tarrow: thaaaanks

outdooracorn (approved):

PR #2713, opened 2026-05-03
Use pull_request_target for mattermost.yml

addshore:
Make the workflow run from the target/base branch instead of PR source branch so that it has access to repo secrets [1].

User provided PRs can not change the workflow code that is running, unless the repo is already compromised.

[1]: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

### DIFF
--- a/.github/workflows/mattermost.yml
+++ b/.github/workflows/mattermost.yml
@@ -1,6 +1,8 @@
 on:
-  pull_request:
+  pull_request_target:
     types: [ opened, reopened ]
+
+permissions: {}
 
 jobs:
   build:


.github/workflows/mattermost.yml
```diff
@@ -1,7 +1,9 @@
 on:
-  pull_request:
+  pull_request_target:
     types: [ opened, reopened ]
 
+permissions: {}
```
outdooracorn, line 5: How come this change?

outdooracorn (commented):
So what does this PR do? Please state that in the PR description.

.github/workflows/mattermost.yml
addshore: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions:~:text=When%20a%20workflow%20is%20triggered%20by%20the%20pull_request_target%20event%2C%20the%20GITHUB_TOKEN%20is%20granted%20read/write%20repository%20permission%2C%20even%20when%20it%20is%20triggered%20from%20a%20public%20fork.%20For%20more%20information%2C%20see%20Events%20that%20trigger%20workflows.

.github/workflows/mattermost.yml
addshore, line 12: Arguably another improvement would either be pinning this image, OR just not using the image and calling the webhook directly.
As this action is technically unpinned external code (however this was a problem that already existed)

PR #2717, opened 2026-05-04
Fix: Add helm install/upgrade flags --take-ownership in API skaffold deploy

dati18:
## Describe the changes
- Add helm install/upgrade flags `--take-ownership` in API skaffold deploy to let Helm take over resources managed by Argo

Bug: T425368

### DIFF
--- a/skaffold/skaffold.yaml
+++ b/skaffold/skaffold.yaml
@@ -87,6 +87,11 @@ profiles:
     deploy:
       kubeContext: minikube-wbaas
       helm:
+        flags:
+          install:
+            - --take-ownership
+          upgrade:
+            - --take-ownership
         releases:
           - name: api
             chartPath: ./../../charts/charts/api


skaffold/skaffold.yaml
outdooracorn, line 1: Oooo, I'm excited about trying this out! 👀 

Two quick thoughts before looking at this properly:
* How do you get back to a local dev env that isn't modified by `skaffold`? Is it as simple as doing an Argo Sync?
* If this works as expected, we should do this to both the apps managed by Argo (i.e `api` and `ui`)

dati18: - yes, you can do Argo Sync, or just checkout any state of the api repo and... skaffold again :P
- that's the neat part, there is no problem `skaffolding` ui and I don't know why...

tarrow: Well, in a super annoying turn of events github lost my long comment here. I'll try and write a terser version:

- this works
- caveat: I had to do something manual to get valid `.tmp.values.api.yaml`/`.tmp.values.ui.yaml` - follow up required but @dang is already looking into one solution in #2718
- running argo sync will return the workloads to a similar state
- it does not remove the helm ownership labels:
```
    meta.helm.sh/release-name: ui
    meta.helm.sh/release-namespace: default
```
- it does also not remove the helm secrets that are used to track helm releases
- I expect this level of dirty cluster state is fine
- I confirmed that the specific issue with `api` is the presence of `serviceaccount/api-defaultrole` which cannot be taken over by helm without this flag
- `ui` doesn't have any `serviceaccount` Resources so there is no issue
- therefore I think if we should add this flag also to `ui` is debatable
- to not let perfect be the enemy of the good I'll approve and merge this now

PR #2718, opened 2026-05-05
Fix: fix helmfile-values script to generate correct temp files

dati18:
- In the `skaffold/helmfile-values` script:
  - `api` and `ui` are now handled through `./bin/generate-values` instead of `helmfile write-values` like other releases
  - Add `#!/usr/bin/env bash` and `set -euo pipefail, explicit -h` to tighten the script
  - Add `cat "$OUTPUT_YAML"` instead of rebuilding temp path string manually
- In the `bin/generate-values`:
  - Add `set -euo pipefail` and safer CI detection with `${CI:-}`
  - Resolve the Helmfile source path from the repo root
  - Use `.yaml.gotmpl` suffix for the temp Helmfile file before being passed to `helmfile --file`

Bug: T425426

### DIFF
--- a/bin/generate-values
+++ b/bin/generate-values
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-if [[ -n "${CI}" ]]; then
+set -euo pipefail
+
+if [[ -n "${CI:-}" ]]; then
     set -x
 fi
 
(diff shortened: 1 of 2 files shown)


bin/generate-values
```diff
@@ -15,8 +17,8 @@ RELEASE="$2"
 
 ROOT=$(git rev-parse --show-toplevel)
 OUTPUT_TEMPLATE="${ROOT}/k8s/argocd/${ENVIRONMENT}/${RELEASE}.values.yaml"
-HELMFILE="k8s/helmfile/only-for-argo-value-generation.yaml"
-TMP_HELMFILE="$(dirname ${HELMFILE})/.tmp_helmfile.$(mktemp -u XXXXXX).yaml"
+HELMFILE="${ROOT}/k8s/helmfile/only-for-argo-value-generation.yaml"
+TMP_HELMFILE="$(dirname "${HELMFILE}")/.tmp_helmfile.$(mktemp -u XXXXXX).yaml.gotmpl"
 
 if [[ -n "$3" ]]; then
```
deer-wmde, line 23: currently when running the script with only 2 arguments I get `./bin/generate-values: line 23: $3: unbound variable `

since `set -u` got added in this PR, we can't use `$3` directly here anymore as it is an optional argument.

we could add something like `USER_OUTPUT_TEMPLATE="${3:-}"` to line 17 and use that var then instead of `$3`

skaffold/helmfile-values
```diff
@@ -40,23 +42,34 @@ while getopts ":r:e:hn" option; do
          TARGET_ENVIRONMENT=$OPTARG;;
       n)
          REMOVE_IMAGE_YAML=1;;
+      h) # display Help
```
deer-wmde, line 45: this case is not needed as the one below (`*`) will also match `h`

PR #2735, opened 2026-05-08
Add `argo-sync-api` command to Makefile

dati18:
Add `argo-sync-api` command to Makefile so that tab-completion works


outdooracorn (approved):
Thanks :shipit:

PR #2739, opened 2026-05-08
shellcheck: streamline scripts

deer-wmde:
- refactors bash scripts to comply with shellcheck linting
- contains also small fixes and improvements found during review

https://phabricator.wikimedia.org/T302394

Bug: [T302394](https://phabricator.wikimedia.org/T302394)

### DIFF
--- a/bin/generate-values
+++ b/bin/generate-values
@@ -13,8 +13,7 @@ function usage() {
 ENVIRONMENT="$1"
 RELEASE="$2"
 
-# shellcheck disable=SC2046,SC2128,2086
-ROOT=$(realpath $(dirname $(realpath $BASH_SOURCE))/..) # absolute path of the wbaas-deploy repository
+ROOT=$(git rev-parse --show-toplevel)
 OUTPUT_TEMPLATE="${ROOT}/k8s/argocd/${ENVIRONMENT}/${RELEASE}.values.yaml"
 HELMFILE="k8s/helmfile/only-for-argo-value-generation.yaml"
 TMP_HELMFILE="$(dirname ${HELMFILE})/.tmp_helmfile.$(mktemp -u XXXXXX).yaml"
--- a/bin/generate-values
+++ b/bin/generate-values
@@ -6,15 +6,15 @@ fi
 
 function usage() {
     echo
-    echo "usage: $(basename $0) <environment> <release-name> [output-file-template]"
+    echo "usage: $(basename "$0") <environment> <release-name> [output-file-template]"
     echo
 }
 
 ENVIRONMENT="$1"
 RELEASE="$2"
 
-# absolute path of the wbaas-deploy repository
-ROOT=$(realpath $(dirname $(realpath $BASH_SOURCE))/..)
+# shellcheck disable=SC2046,SC2128,2086
+ROOT=$(realpath $(dirname $(realpath $BASH_SOURCE))/..) # absolute path of the wbaas-deploy repository
 OUTPUT_TEMPLATE="${ROOT}/k8s/argocd/${ENVIRONMENT}/${RELEASE}.values.yaml"
 HELMFILE="k8s/helmfile/only-for-argo-value-generation.yaml"
 TMP_HELMFILE="$(dirname ${HELMFILE})/.tmp_helmfile.$(mktemp -u XXXXXX).yaml"
--- a/bin/helmfile-env-diff
+++ b/bin/helmfile-env-diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
 Help()
 {
@@ -32,7 +32,7 @@ while getopts ":e:f:r:h" option; do
          ENVIRONMENT2=$OPTARG;;
       r)
          RELEASE=$OPTARG;;
-      h) # display Help
+      *) # display Help
          Help
          exit;;
    esac
@@ -42,5 +42,5 @@ SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 HELMFILE_DIR="$SCRIPT_DIR/../k8s/helmfile"
 
 echo "RED is $ENVIRONMENT1, GREEN is $ENVIRONMENT2"
-diff $HELMFILE_DIR/env/$ENVIRONMENT1/$RELEASE.values.yaml.gotmpl $HELMFILE_DIR/env/$ENVIRONMENT2/$RELEASE.values.yaml.gotmpl  --color
+diff "$HELMFILE_DIR/env/$ENVIRONMENT1/$RELEASE.values.yaml.gotmpl" "$HELMFILE_DIR/env/$ENVIRONMENT2/$RELEASE.values.yaml.gotmpl"  --color
 echo "RED is $ENVIRONMENT1, GREEN is $ENVIRONMENT2"
--- a/bin/list-wiki-domains
+++ b/bin/list-wiki-domains
@@ -1,4 +1,4 @@
 #!/bin/bash
 
+# shellcheck disable=SC2016
 kubectl exec -ti deployments/api-app-backend -- php artisan tinker --execute 'foreach(App\Wiki::get() as $wiki): echo "$wiki->domain" . "\n"; endforeach'
-
--- a/bin/list-wiki-domains
+++ b/bin/list-wiki-domains
@@ -1,4 +1,5 @@
 #!/bin/bash
 
+# PHP code is intentionally wrapped in single quotes to prevent being interpreted by the shell
 # shellcheck disable=SC2016
 kubectl exec -ti deployments/api-app-backend -- php artisan tinker --execute 'foreach(App\Wiki::get() as $wiki): echo "$wiki->domain" . "\n"; endforeach'
--- a/bin/local/create-local-wiki.sh
+++ b/bin/local/create-local-wiki.sh
@@ -48,13 +48,13 @@ LOGIN_MAIL=$(echo "${LOGIN_RESPONSE}" | jq -r '.user.email')
 [[ "$LOGIN_MAIL" == "$USER_MAIL" ]] || { echo "error: login failed." 1>&2; exit 1; }
 
 ## Create wiki
-# shellcheck disable=SC2034
 CREATE_WIKI_JSON_DATA=$(jo -- \
     domain="${USER_WIKI_DOMAIN}" \
     sitename="${USER_WIKI_NAME}" \
     username="${USER_WIKI_ADMIN}" \
-    -s profile="$(jo purpose=decide_later temporality=decide_later)") \
-    knowledgeEquityResponse="$(jo selectedOption="unsure" freeTextResponse="")"
+    -s profile="$(jo purpose=decide_later temporality=decide_later)" \
+    knowledgeEquityResponse="$(jo selectedOption="unsure" freeTextResponse="")")
+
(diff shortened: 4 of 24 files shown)


bin/local/create-local-wiki.sh
outdooracorn: I think shellcheck picked up on something wrong here. The `knowledgeEquityResponse` line isn't part of the `CREATE_WIKI_JSON_DATA` subshell. You can see if you do `echo $CREATE_WIKI_JSON_DATA; exit;`, `knowledgeEquityResponse` isn't included:
```
$ ./bin/local/create-local-wiki.sh 
Current kube context: 'minikube-wbaas'
{"domain":"local-test-wiki.wbaas.dev","sitename":"Local Test Wiki","username":"Admin","profile":"{\"purpose\":\"decide_later\",\"temporality\":\"decide_later\"}"}
```

The last closing brace `)` on line 56, should actually be on line 57.

```
## Create wiki
CREATE_WIKI_JSON_DATA=$(jo -- \
    domain="${USER_WIKI_DOMAIN}" \
    sitename="${USER_WIKI_NAME}" \
    username="${USER_WIKI_ADMIN}" \
    -s profile="$(jo purpose=decide_later temporality=decide_later)" \
    knowledgeEquityResponse="$(jo selectedOption="unsure" freeTextResponse="")")
```

bin/generate-values
```diff
@@ -6,15 +6,15 @@ fi
...
     echo
 }
 
 ENVIRONMENT="$1"
 RELEASE="$2"
 
-# absolute path of the wbaas-deploy repository
-ROOT=$(realpath $(dirname $(realpath $BASH_SOURCE))/..)
+# shellcheck disable=SC2046,SC2128,2086
+ROOT=$(realpath $(dirname $(realpath $BASH_SOURCE))/..) # absolute path of the wbaas-deploy repository
```
outdooracorn: How come disabling these checks rather than fixing them?

This seems to work for me and shellcheck doesn't complain:
```suggestion
# absolute path of the wbaas-deploy repository
ROOT=$(realpath "$(dirname "$(realpath "$0")")/..")
```

disclaimer: I only tested this by `echo`ing the actual output next to the expected output:
```
echo "ROOT: $ROOT";
echo "COPY: /home/ollie/git/github/outdooracorn/wbaas-deploy";
exit;
```

---

Alternatively, we could use the git command `git rev-parse --show-toplevel` to get the root directory path. This command has the benefit of working even if we move where this script is located.

bin/helmfile-env-diff
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
```
outdooracorn, line 1: Interesting! Did shellcheck request this change? Got any reading material as to why one is better to use over the other?

bin/helmfile-env-diff
```diff
@@ -42,5 +42,5 @@ SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 HELMFILE_DIR="$SCRIPT_DIR/../k8s/helmfile"
 
 echo "RED is $ENVIRONMENT1, GREEN is $ENVIRONMENT2"
-diff $HELMFILE_DIR/env/$ENVIRONMENT1/$RELEASE.values.yaml.gotmpl $HELMFILE_DIR/env/$ENVIRONMENT2/$RELEASE.values.yaml.gotmpl  --color
+diff "$HELMFILE_DIR/env/$ENVIRONMENT1/$RELEASE.values.yaml.gotmpl" "$HELMFILE_DIR/env/$ENVIRONMENT2/$RELEASE.values.yaml.gotmpl"  --color
 echo "RED is $ENVIRONMENT1, GREEN is $ENVIRONMENT2"
```
outdooracorn, line 46: I was surprised that shellcheck didn't complain about missing newline at the end of file, but it only complains if it is missing the trailing linefeed https://github.com/koalaman/shellcheck/issues/1334.

Posting comment just as an FYI and that we may want to look at using something like [shfmt](https://github.com/mvdan/sh) as well.

bin/list-wiki-domains
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
+# shellcheck disable=SC2016
```
outdooracorn, line 4: I think it would be good to include a reason for why we are disabling this check.

```suggestion
# PHP code is intentionally wrapped in single quotes to prevent being interpreted by the shell
# shellcheck disable=SC2016
```

bin/reset-argocd-password
```diff
@@ -11,6 +11,7 @@ function patch_password() {
 PASSWORD=$(tr -dc A-Za-z0-9 </dev/urandom | head -c 32)
 
 # https://github.com/argoproj/argo-helm/blob/93b7a946503d2b0a7cfe09ac619d91e7c9d16c83/charts/argo-cd/values.yaml#L532
+# shellcheck disable=SC2016
```
outdooracorn, line 15: I think it would be good to include a reason for why we are disabling this check.

```suggestion
# sed argument is intentionally wrapped in single quotes to prevent being interpreted by the shell
# shellcheck disable=SC2016
```

bin/update-ingress-service
```diff
@@ -20,11 +20,11 @@ INGRESSES=$(kubectl get ingress -l wbstack-ingress-generation -o json)
 
 echo "$INGRESSES" > "wbstack-ingresses-$(date +%s).json.bak"
 
-UPDATED_INGRESSES=$(echo "$INGRESSES" | jq '.items[].spec.rules[].http.paths[].backend.service.name="'$SERVICE'"')
+UPDATED_INGRESSES=$(echo "$INGRESSES" | jq '.items[].spec.rules[].http.paths[].backend.service.name="'"${SERVICE}"'"')
```
outdooracorn: What do you think about using the `--arg` argument of `jq` instead?
https://stackoverflow.com/a/34747439/708560
Might be easier to read than all of these quotes 😅

skaffold/helmfile-values
```diff
@@ -51,12 +51,12 @@ OUTPUT_YAML=$SCRIPT_DIR/.tmp.values.$TARGET_RELEASE.yaml
...
 if [[ "$REMOVE_IMAGE_YAML" == 1 ]]; then
-   yq eval 'del(.image)' --no-colors $OUTPUT_YAML > $OUTPUT_YAML.altered
-   mv $OUTPUT_YAML.altered $OUTPUT_YAML
+   yq eval 'del(.image)' --no-colors "$OUTPUT_YAML" > "$OUTPUT_YAML".altered
+   mv "$OUTPUT_YAML".altered "$OUTPUT_YAML"
 fi
 
-cat $SCRIPT_DIR/.tmp.values.$TARGET_RELEASE.yaml
\ No newline at end of file
+cat "$SCRIPT_DIR"/.tmp.values."$TARGET_RELEASE".yaml
```
outdooracorn: Optional nitpick: how about quoting the whole string rather than just the variables? I find this easier to read.

```suggestion
cat "$SCRIPT_DIR/.tmp.values.$TARGET_RELEASE.yaml"
```

bin/queryservice-delete-dump-timestamp-in-namespaces
```diff
@@ -3,10 +3,10 @@
 # This script requires you to pass a file containing a list of the namespaces as the first argument
 # This scripts expects the port of the queryservice to be forwarded to the host machine
 
-while read namespace; do
+while read -r namespace; do
 
     DELETE_QUERY="DELETE WHERE {<http://wikiba.se/ontology#Dump> <http://schema.org/dateModified> ?time}"
-    RESULT=$(curl -X POST http://localhost:9999/bigdata/namespace/$namespace/sparql/ --silent --data-raw "update=$DELETE_QUERY" -H 'Accept: application/sparql-results+json')
+    RESULT=$(curl -X POST http://localhost:9999/bigdata/namespace/"$namespace"/sparql/ --silent --data-raw "update=$DELETE_QUERY" -H 'Accept: application/sparql-results+json')
```
outdooracorn: Optional nitpick: how about quoting the whole string rather than just the variables? I find this easier to read.

```suggestion
    RESULT=$(curl -X POST "http://localhost:9999/bigdata/namespace/$namespace/sparql/" --silent --data-raw "update=$DELETE_QUERY" -H 'Accept: application/sparql-results+json')
```

outdooracorn (dismissed):
Looking good! I spotted a few issues, improvements, and optional nitpicks.

EDIT: I hope this was still in draft because of the merge chain, and not because you didn't want anyone to review it yet :see_no_evil:

bin/helmfile-env-diff
deer-wmde: it did not - if anything using `/usr/bin/env` should be more compatible as far as I now. It was just less effort to change this one, as the alternative would be imo to change all scripts to use `env`

I also have nothing against using `env`, I just think we should do one option consistently

bin/queryservice-delete-dump-timestamp-in-namespaces
deer-wmde: yes! I wanted to revise all changes again because too late I realized shellcheck does not "need" what it suggests (meaning, not in exactly the same way it suggests, and some alternatives are much easier to read)

bin/generate-values
deer-wmde: yeah, previously I just applied what shellcheck suggested and it made that line so difficult to read. As pointed out in the other comment, I only later realized it doesnt strictly "need" it in the way it suggests.

Thanks for the alternative command!

bin/update-ingress-service
deer-wmde: fancy! good idea!

bin/helmfile-env-diff
deer-wmde: Thanks for suggesting that tool! I was missing something like that from shellcheck and I think something like that would be benefical as well. I'll make a new PR for that

bin/local/create-local-wiki.sh
deer-wmde: oooh great catch!!

bin/generate-values
tarrow: looks like we decided to use `git rev-parse --show-toplevel` and there are now no exceptions.

bin/helmfile-env-diff
outdooracorn: Thanks for the context.

> I also have nothing against using env, I just think we should do one option consistently

Agreed!

No further comments from me on this. :)

tarrow: Very quick casual bit of googling:
https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash

tl;dr - using env makes the script more portable and search for a version of bash the user might have in their environment (e.g. maybe they have `~/bin/bash` in their PATH) where as the full, absolute path provides no flexibility.

In our use case we want to use the absolute path. It makes it less likely an in advised `bash` somewhere in someone's PATH will trip us up and we're not worried about cleverly negotiating to a specific, non-globally installed bash version.

skaffold/helmfile-values
tarrow: @outdooracorn says below he's happy that his comments have been address. I think we can follow up on this nitpick at a later date if wanted

bin/queryservice-delete-dump-timestamp-in-namespaces
tarrow: @outdooracorn says below he's happy that his comments have been address. I think we can follow up on this nitpick at a later date if wanted

tarrow (approved):

PR #2765, opened 2026-05-23
Use ingress-nginx from the chainguard fork built by porter

iMartyn:
[porter](https://github.com/porter-dev/ingress-nginx) are building free images of chainguard's fork of ingress-nginx that chainguard are keeping behind their paywall.  This is not a longterm solution but better than sticking on the old one.

### DIFF
--- a/k8s/helmfile/helmfile.yaml.gotmpl
+++ b/k8s/helmfile/helmfile.yaml.gotmpl
@@ -114,6 +114,15 @@ releases:
     # that can be used instead of YAML anchors. When upgrading helmfile, check
     # if we can use this feature here instead.
     <<: *default_release
+    set:
+    - name: controller.image.registry
+      value: ghcr.io
+    - name: controller.image.image
+      value: porter-dev/ingress-nginx-controller
+    - name: controller.image.tag
+      value: v1.15.5
+    - name: controller.image.digest
+      value: null
 
   - name: sql
     namespace: default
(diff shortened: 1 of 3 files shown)


tarrow (approved):
trying an approve to see if that will allow status checks to run?

k8s/helmfile/helmfile.yaml.gotmpl
```diff
@@ -114,6 +114,15 @@ releases:
     # that can be used instead of YAML anchors. When upgrading helmfile, check
     # if we can use this feature here instead.
     <<: *default_release
+    set:
```
tarrow: if we want to inline the value here I think we need to use `values` not `set` which I think is only for templated values see: https://helmfile.readthedocs.io/en/latest/configuration/#full-reference

tarrow (changes requested):
I think we want to make sure we can deploy this to staging before production.

The easiest way to do this is to follow the pattern that's currently there and put this new set of values in the per service values file. i.e. https://github.com/wmde/wbaas-deploy/blob/main/k8s/helmfile/env/staging/ingress-nginx.values.yaml.gotmpl

dati18 (dismissed):
approve to run checks, please don't merge

dati18 (dismissed):
aprrove to test again

k8s/helmfile/helmfile.yaml.gotmpl
dati18: The output is as expected on minikube. @tarrow Do you want to check again and slap an `approve` on this?

```
$ kubectl get pod -n kube-system -l app.kubernetes.io/name=ingress-nginx,app.kubernetes.io/component=controller -o json | jq '.items[].spec.containers[0].image'
"ghcr.io/porter-dev/ingress-nginx-controller:v1.15.5"
```

dati18 (approved):

tarrow (approved):
LGTM - let's try it on staging

PR #2966, opened 2026-08-10
production+staging(quickstatements): increase memory resource

tarrow:
Increase (triple) request and limit of memory to try and prevent being killed by OOM.

This commit increases the resources in both environments since we don't adjust them in staging and considering the limited size of the resources it doesn't seem worth having the disparity vs increased consumption.

Bug: T434076


ranyardm-wmde (approved):
Seems like a good thing to try.

PR #2977, opened 2026-08-18
Add sending reverification emails to user-administration.md docs

outdooracorn:
Update the User Administration docs to include a "Sending a reverification email" section.

Bug: T435171


tarrow (changes requested):
I think it would be good to add *something* about how this differs from the user triggering their own revalidation email and when this should be done versus us manually intervening.

Isn't this exactly the same job and method of calling that happens when the user clicks resend @ https://github.com/wbstack/api/blob/29226a07f28c1171ec7cd0ac294c68eb943a1644/app/Http/Controllers/UserVerificationTokenController.php#L52 ?

PR #2982, opened 2026-08-24
prod: increase backup scratch disk size

deer-wmde:
As the total size of the backups increase, the scratch disk space needs adjustment too from time to time.
Last time[1] we went from 128GB to 256GB.
In this PR I doubled it again to 512GB, which is maybe a bit much, but also not too dramatic for this temporary disk I think.

[1] https://github.com/wmde/wbaas-deploy/pull/2102

https://phabricator.wikimedia.org/T435042

Bug: T435042


tarrow (approved):
yeah, doubling each time is eventually going to be silly but for now I think it's OK

